Issue #5, opened 2018-07-20
Why do you fork the yaml?


Wing924:
I want to know the difference between forked yaml and original one.

l2isbad:
[field validation added](https://github.com/l2isbad/yaml/commit/aa9c5dfde78ee8088f1dc6a4b0b7c175ec35f863)
 (new field tags: `required`, `inrange`, `inregex(p|d)`)

```go
// inregexp returns error if not matches
// inregexd returns error if matches
type Example struct {
	Range    int    `yaml:"filed1,inrange:[1:10]"` // range:[:10] || range:[1:]
	Regexp   string `yaml:"field2,inregexp:REGULAR_EXPRESSION"`
	Required bool   `yaml:"field3,required"`
}

```

Wing924:
If we really need this, how about send that PR to original one?
Use the forked library will much increase maintenance cost.

l2isbad:
> If we really need this

Yes, we need this. It's a very nice feature. 

>  how about send that PR to original one?

Well, i think this is very specific changes which will not be accepted.

> Use the forked library will much increase maintenance cost.

How?

Wing924:
I think this is what you want:
<https://github.com/go-playground/validator>

sample:
```go
// User contains user information
type User struct {
	FirstName      string     `validate:"required"`
	LastName       string     `validate:"required"`
	Age            uint8      `validate:"gte=0,lte=130"`
	Email          string     `validate:"required,email"`
	FavouriteColor string     `validate:"iscolor"`                // alias for 'hexcolor|rgb|rgba|hsl|hsla'
	Addresses      []*Address `validate:"required,dive,required"` // a person can have a home and cottage...
}
```

I suggest not use forked yaml and use standard yaml with validator instead.

l2isbad:
This package is nice, looks like it is suitable for our needs. 
Give me some time to rewrite `validation` loginc

l2isbad:
Ok, we are free from my `yaml` fork :smile:

PR #6, opened 2018-07-20
use dep to manage the dependency

Wing924:
what I do
-------------

```
dep init
dep prune
```

what I not do
-----------------

Change any source codes.

<https://github.com/golang/dep>


Wing924:
@l2isbad Could you check this? Without this, other people may not be able to compile it from same source.

l2isbad:
@Wing924 
i think #5 should be resolved first. I mean you need to re vendor after.

Wing924:
@l2isbad I updated the dep after your commits.

PR #13, opened 2018-07-24
Support prometheus metrics

Wing924:
status:
- Support prometheus metrics
  - Coding, unit test: done
- Support Springboot
  - Coding: POC level
  - unit test: not yet
  - manual test: done


l2isbad:
Do we realy need all these 10k+ lines of dependencies? Looks a lot

l2isbad:
Hm, only 1 chart with 2 dimensions? :smile:

internal/modules/springboot2/springboot2.go
```diff
@@ -0,0 +1,56 @@
...
+// Check Check
+func (s *Springboot2) Check() bool {
+	s.prom = prometheus.New(s.Client.CreateHttpClient(), s.Request)
+	metrics, err := s.prom.GetMetrics()
+	if err != nil {
+		s.Error(err)
+		return false
+	}
+	jvmMemory := metrics.FindByName("jvm_memory_used_bytes")
+	return len(jvmMemory) > 0
```
l2isbad: `*raw.Charts` should be added to `cooked.Charts` (`AddMany`) in `Check`

internal/modules/springboot2/charts.go
```diff
@@ -0,0 +1,19 @@
+package springboot2
+
+import (
+	"github.com/l2isbad/go.d.plugin/internal/pkg/charts/raw"
+)
+
+var charts = raw.Charts{
```
l2isbad: I suggest to use aliases.
```go
type (
	Charts      = raw.Charts
	Order       = raw.Order
	Definitions = raw.Definitions
	Chart       = raw.Chart
	Options     = raw.Options
	Dimensions  = raw.Dimensions
	Dimension   = raw.Dimension
)
```

Much cleaner

Wing924:
> Do we realy need all these 10k+ lines of dependencies? Looks a lot
To be honest i am a bit worried about that.

I think it's not a problem. please ignore `/vendor` directory. Everything is generate by `dep`.

> Hm, only 1 chart with 2 dimensions?

Yes, springboot part is just POC for prometheus helper. I'll update in another PR later once the plugin API reach stable.

internal/modules/springboot2/charts.go
Wing924: I used to use aliases, but `go lint` complains write comment :(

l2isbad: tune it then :smile: , **aliases are nice**

l2isbad: or ignore in that particular case

Wing924: I wonder why should we need both `raw` and `cooked` version?

Wing924: > if it will looks bad - no go (like map[string]*Chart, even if it makes sense)

I agree.

> ignore in that particular case

The green wavy line (go lint) is very harmful for my eye and there no way to disable it.

l2isbad: I know that feel bro 
Does bunch of `raw.*`'s is not harmful for you? :smiley: Especially if you know that they can be avoided... Do alias, remove raw, save the kittens

l2isbad: Just joking. Let's leave it as is. I will suffer, i can handle it.

l2isbad:
does it ready?

Wing924:
> *raw.Charts should be added to cooked.Charts (AddMany) in Check

I haven't fix it. I'll fix it tomorrow.

l2isbad:
> I think it's not a problem. 

more binary size, more memory usage in runtime. I think we should check before/after adding new external dependency.

I think if functionality is simple we can implement it ourselves (i did it for cmd args parser), the whole validator package is not needed too, it can be relativly easy replaced by our own implementation.

Am i wrong here?

l2isbad:
> Have you ever tried flag package?

I have.
The reason I did not use it is lack of short names.

https://github.com/l2isbad/go.d.plugin/blob/5609041085e0063fdff64decf7874999236b0521/internal/pkg/cli/cli.go#L26-L27

`./go.d.plugin -m module_name -d` > `./go.d.plugin -module module_name -debug`

Wing924:
@l2isbad 
If you just need simple, standard flag is enough. Short name function is not a must level requirement because this program will be called from netdata, not human.

If you want the code looks cool, I think this flag library looks nice and widely used.
<https://github.com/jessevdk/go-flags>

l2isbad:
> will be called from netdata, not human.

debug mode for humans, i use it a lot during modules development.

l2isbad:
https://github.com/l2isbad/go.d.plugin/blob/8735f4c6254e974d24c7110429072b1acd7791f2/internal/pkg/helpers/prometheus/prometheus.go#L94-L95

Don't we need here to read the whole body to reuse the connection?

l2isbad:
like
https://github.com/l2isbad/go.d.plugin/blob/8735f4c6254e974d24c7110429072b1acd7791f2/internal/modules/httpcheck/httpcheck.go#L124-L127

Wing924:
> Don't we need here to read the whole body to reuse the connection?

Good point! I'll check it.

l2isbad:
Did you check it?

Wing924:
I’ll send PR ASAP

l2isbad:
yes, better to do it, just to not forget

PR #16, opened 2018-07-26
charts refactor

### LINKED ISSUE #14: cooked package objects structure
**charts** current implementation:

```go
type charts struct {
	charts   map[string]*Chart 
	bc       baseConfHook
	priority int
}
```

I suggest to change to

```go
type charts struct {
	charts  []*Chart  // !! changes
	bc       baseConfHook
	priority int
}
```

**chart** current implementation:
```go
type (
	Chart struct {
        ....
	}
	dimensions []*dimension
	variables  map[string]*variable
)
```

I suggest to change to

```go
type (
	Chart struct {
        ....
	}
	dimensions []*dimension
	variables  []*variable // !! changes
)
```

Feels more natural to me. Duck maps. What do you think @Wing924 ?

### LINKED ISSUE #15: Charts definitions code style
I think this style is simpler and more readable.
How do you think? @l2isbad 
```go
charts.Charts{
    Items: []charts.Chart{
        {
            ID:      "response status",
            Options: charts.Options{Title: "response status", Units: "requests/s", Family: "response"},
            Dimensions: []charts.Dimension{
                {ID: "1xx", Name: "1xx", Algorithm: "incremental"},
                {ID: "2xx", Name: "2xx", Algorithm: "incremental"},
                {ID: "3xx", Name: "3xx", Algorithm: "incremental"},
                {ID: "4xx", Name: "4xx", Algorithm: "incremental"},
            },
        },
        {
            ID:      "response time",
            Options: charts.Options{Title: "response time", Units: "second", Family: "response"},
            Dimensions: []charts.Dimension{
                {ID: "min", Name: "1xx", Divisor: 1000},
                {ID: "max", Name: "2xx", Divisor: 1000},
                {ID: "avg", Name: "3xx", Divisor: 1000},
            },
        },
    },
}
```

l2isbad:
fixes #14 
fixes #15


l2isbad:
This PR: 
- chart, dimensions are simple structs, slice of structs with a few methods.
- proxy charts, dims are wrappers around charts, structs with additional fields, methods which are needed internally

what do you think? @Wing924

internal/godplugin/job/job.go
```diff
@@ -7,20 +7,21 @@ import (
...
 	}
 }
 
 type Job struct {
 	modules.Module
-	*Config
-	timers
+	t       timers
+	C       Config
+	W       *WrappedCharts
```
Wing924: I think full name is more readable.
```
timers       timers
Config       Config
Charts       *WrappedCharts
```

internal/godplugin/job/job.go
```diff
@@ -30,20 +31,20 @@ Done:
...
 
 		sleep := j.nextIn()
-		j.Debugf("sleeping for %s to reach frequency of %d sec", sleep, j.UpdEvery)
+		j.Debugf("sleeping for %s to reach frequency of %d sec", sleep, j.C.UpdEvery)
 		time.Sleep(sleep)
 
-		j.curRun = time.Now()
-		if !j.lastRun.IsZero() {
-			j.sinceLast.Duration = j.curRun.Sub(j.lastRun)
+		j.t.curRun = time.Now()
```
Wing924: https://golang.org/pkg/time/#Ticker 
I think use Ticker rather than timers is simpler and Go style.
We can refactor it in another PR :)

internal/godplugin/job/wrapped_chart.go
```diff
@@ -0,0 +1,211 @@
...
+		p.dimensions = append(p.dimensions, newWrappedDim(v))
+	}
+	for _, v := range chart.Vars {
+		p.variables = append(p.variables, &wrappedVar{v})
+	}
+	return p
+}
+
+type wrappedChart struct {
+	item          charts.Chart
```
Wing924: `chart *charts.Chart`
I think ptr is better because plugin may change the chart.

internal/godplugin/job/wrapped_chart.go
```diff
@@ -0,0 +1,211 @@
...
+
+// ------------------------------------------------------------------------------------
+type chartFlags struct {
+	push      bool
+	created   bool
+	updated   bool
+	obsoleted bool
+}
+
+func (f *chartFlags) setPush(b bool) {
```
Wing924: Do we need all setters?
Just `flags.created = true` is OK

Wing924:
Btw, how about this design?
```go
type Chart {
  // public members, user can set it directly
  Title string
  Dims []Dim
  ...

  // private members for cache, index, etc.
  pushed bool
  created bool
  dimMap map[string]*Dim
}


func (c *Chart) Refresh() {
  // Clear cache, re-index, etc.
}
```

So that we can also remove the `wrappedChart`, `wrappedDim`, etc.

internal/godplugin/job/wrapped_chart.go
l2isbad: No now. It was helpful tho - different color. These flags are super important, so I wanted to see where they are changing

Wing924: understood

internal/godplugin/job/wrapped_chart.go
l2isbad: > plugin may change the chart.

```go
func (w *wrappedChart) SetID(s string) {
    w.item.ID = s
}
```

internal/godplugin/job/job.go
l2isbad: Yeah, different story. Actually we have to use monotonic to calc next run and real time to measure  duration `sinceLast`

internal/godplugin/job/job.go
l2isbad: Ofc it's more readable. `timers` and `Config` should be embedded.

l2isbad:
> So that we can also remove the wrappedChart, wrappedDim, etc.

Why? this functionality belongs to job package. pkg/charts is not for maintaining charts in runtime

l2isbad:
Purpose of the `pkg/charts` is to create charts before running the job.

Wing924:
My design
---------------------

Charts have 2 functions:
* hold data (charts have many chart, chart have many dim, etc.)
* hold state (charts is added? etc.)

Plugins manage the holding data (create/update/delete charts)
Jobs manage the state

Once Plugin updated the chart data, chart notify the job to change state.

l2isbad:
Lets agree on terms.
Plugin - go.d.plugin. His task is to create jobs and run them, wait, then die if there is no active jobs. Nothing more. ?

https://github.com/l2isbad/go.d.plugin/blob/8735f4c6254e974d24c7110429072b1acd7791f2/internal/godplugin/plugin.go#L67-L69

Module. Gathers data. Nothing more. ?

> Plugins manage the holding data (create/update/delete charts)

Do you mean modules? or what?

Wing924:
Sorry, my words is not correct.

My Plugins = web_log, httpcheck, portcheck, etc.

l2isbad:
> Plugins manage the holding data (create/update/delete charts)
> Jobs manage the state

That'is kinda how it works right now. Plugin adds/updates charts (unlimited befrore run, limited after). Job manages the state via chart flags.

l2isbad:
I want full controll on which methods/fields are available for module/job (different kind of sets). This is not possible only by using export/unexport feature, this is why i stick to interfaces.

Wing924:
> I want full controll on which methods/fields are available for module/job (different kind of sets). This is not possible only by using export/unexport feature, this is why i stick to interfaces.

How about use observer pattern?
If module changed charts, notify the job

l2isbad:
Why not. Let's see how it will be. I will do it until sunday

internal/godplugin/job/observer.go
```diff
@@ -0,0 +1,89 @@
...
+	c.push = true
+	if c.obsoleted {
+		c.retries = 0
+		c.updated = false
+		c.obsoleted = false
+	}
+}
+
+// FIXME: DUPLICATE CHARTS, DIMS
+func NewObserver(ch *charts.Charts, hook baseConfHook) *observer {
```
Wing924: Please return a public interface or change func name to `newObserver`

l2isbad:
I think it is much better now @Wing924 What do you think?

Wing924:
```
--- FAIL: TestCharts_AddChart (0.00s)
	charts_test.go:34: excpected 2 charts, but got 1
FAIL
FAIL	github.com/l2isbad/go.d.plugin/internal/pkg/charts	0.016s
```
@l2isbad Could you fix the test?

l2isbad:
Oh, my bad :cry: will do later, on mobile atm

l2isbad:
found some bugz

l2isbad:
@Wing924 any comments?

Wing924:
@l2isbad I'm still reading the code. But it looks much better :)

l2isbad:
Well, it's -2k lines and same functionality lol.

l2isbad:
> The job part is complex

Yes. Job part is not accomplished and, i am sure, has some bugs. I still have to write a lot of test for it (this only applies to chart management in runtime, adding, deleting, updating), but i want to switch to another task :smile:

PR #20, opened 2018-08-01
Fix springboot2 connection leak etc.

Wing924:
Fix springboot2 connection leak #13 
Fix StrToMap panic if s is not a ptr
Fix logger panic if logger is nil

### DIFF
--- a/internal/pkg/utils/funcs.go
+++ b/internal/pkg/utils/funcs.go
@@ -7,7 +7,15 @@ const fieldTagName = "stm"
 // StrToMap Convert struct to map[string]int64 based on stm tag
 func StrToMap(s interface{}) map[string]int64 {
 	rv := make(map[string]int64)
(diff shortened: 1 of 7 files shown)


Wing924:
@l2isbad it's ready, please check it.

internal/pkg/utils/funcs.go
```diff
@@ -7,7 +7,15 @@ const fieldTagName = "stm"
...
-	t, v := reflect.TypeOf(s).Elem(), reflect.ValueOf(s).Elem()
+	t := reflect.TypeOf(s)
+
+	if t.Kind() == reflect.Ptr {
+		t = t.Elem()
+	}
+	v := reflect.ValueOf(s)
+	if v.Kind() == reflect.Ptr {
+		v = v.Elem()
+	}
```
l2isbad: I suggest to shorten to

```go
	v := reflect.Indirect(reflect.ValueOf(s))
	t := v.Type()
```

l2isbad:
Nice!

PR #22, opened 2018-08-04
user decorator pattern in helpers.web

l2isbad:
Yo @Wing924

have a look


l2isbad:
@Wing924 
I think i was wrong
To reuse same client for different connections (we have no modules that need such functionality, but still) should be

```go
type RawRequest struct {
	URL      string            `yaml:"url" validate:"required,url"`
	Body     string            `yaml:"body"`
	Method   string            `yaml:"method" validate:"isdefault|oneof=GET POST"`
	Header   map[string]string `yaml:"headers"`
	Username string            `yaml:"username"`
	Password string            `yaml:"password"`
}

type RawClient struct {
	FollowRedirect bool           `yaml:"follow_redirects"`
	Timeout        utils.Duration `yaml:"timeout"`
	TLSVerify      bool           `yaml:"tls_verify"`
	ProxyURL       string         `yaml:"proxy_url"`
	ProxyUsername  string         `yaml:"proxy_username"`
	ProxyPassword  string         `yaml:"proxy_password"`
}
```

What do you think?

Wing924:
So you want to split query context and query itself, right?
Because Go also split client and request (Java don't split it), I think this change make sense.
But I don't know if we need such complex one.
Most netdata's requests are to localhost without proxy, https, headers. etc.

If we can keep the conf simple and clean, we can split it; otherwise, keep as is is not a bad idea.

l2isbad:
Ok, i will do a PR tomorrow, let's see how it works out

PR #23, opened 2018-08-07
add msg per interval limit to logger

l2isbad:
@Wing924

### DIFF
--- a/internal/pkg/logger/logger.go
+++ b/internal/pkg/logger/logger.go
@@ -4,11 +4,15 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"time"
 )
 
-var dummy = &Logger{
-	log: log.New(colored{}, "", log.Ldate|log.Ltime),
-}
+var (
+	msgPerInterval = 60
+	interval = time.Second
+)
+
+var dummy = New("dummy", "dummy")
 
 func New(modName, jobName string) *Logger {
 	return &Logger{
@@ -22,6 +26,9 @@ type Logger struct {
 	log     *log.Logger
 	modName string
 	jobName string
+
+	count  int
+	ticker *time.Ticker
 }
 
 func (l *Logger) Critical(a ...interface{}) {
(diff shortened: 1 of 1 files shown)


Wing924:
What does this PR do?

l2isbad:
Adds msg per interval limit to logger.

Wing924:
What do you mean?

Wing924:
Could you tell me the motivation?
I don't understand why we need to limit the log output.

l2isbad:
Netdata writes all error msgs to a file. We have to implement some flood protection mechanism, otherwise any module can flood netdata with error messages.

Example:
We want to log `unmatched` line in `web_log`, but what if there is several of thoushand unmatched lines? Anyway it's a good thing to have.

Wing924:
> We want to log unmatched line in web_log, but what if there is several of thoushand unmatched lines? Anyway it's a good thing to have.

It make sense.

https://github.com/golang/go/wiki/RateLimiting
I think implement it like this is easier and overhead is less.

Wing924:
https://godoc.org/golang.org/x/time/rate

l2isbad:
Well, what do you mean by overhead in my PR?
I don't see any.

internal/pkg/logger/logger.go
```diff
@@ -70,17 +77,45 @@ func (l *Logger) print(level Severity, a ...interface{}) {
...
+// SetModName sets logger modName
 func SetModName(l *Logger, modName string) {
 	l.modName = modName
 }
+
+//TODO: stop ticker, do not hard code msgPerInterval, interval?
+// SetLimit sets logger ticker
+// After that it's not allowed to log more than 60 messages per 1 second.
+func SetLimit(l *Logger) {
+	l.ticker = time.NewTicker(interval)
```
Wing924: we don't need ticker per logger.
a global one is enough.

internal/pkg/logger/logger.go
```diff
@@ -70,17 +77,45 @@ func (l *Logger) print(level Severity, a ...interface{}) {
...
+		return
 	}
+
+	select {
+	case <-l.ticker.C:
+		l.count = 0
+	default:
+	}
+
+	if l.count > msgPerInterval {
```
Wing924: race condition here

internal/pkg/logger/logger.go
```diff
@@ -70,17 +77,45 @@ func (l *Logger) print(level Severity, a ...interface{}) {
 		return
 	}
 
-	if l != nil && l.log != nil {
+	if l == nil || l.log == nil {
+		dummy.log.Printf("go.d: %s: %s: %s: %s", level, l.modName, l.jobName, fmt.Sprintln(a...))
```
Wing924: null ptr reference: `l.modName`, `l.jobName`

internal/pkg/logger/logger.go
```diff
@@ -70,17 +77,45 @@ func (l *Logger) print(level Severity, a ...interface{}) {
...
+		l.count = 0
+	default:
+	}
+
+	if l.count > msgPerInterval {
+		return
+	}
+
+	if sevLevel < DEBUG {
+		l.count++
```
Wing924: not atomic operation => race condition

l2isbad: yes, ofc, my inattention :cry:

l2isbad:
Seem still a race here


```go
	case <-l.resetCh:
		l.count = 0
	default:
	}
```

Should i use `atomic.AddInt64(&l.count, -atomic.LoadInt64(&l.count))` ? lol

l2isbad:
I see `https://godoc.org/golang.org/x/time/rate` uses `time.Now()` for every call. Not sure how it's heavy, but looks like we have to try to avoid it.

Wing924:
How about this:
each logger:
```go
if atomic.AddInt32(&l.count, 1) > 60 {
  return
}
// print log
```
timer:
```go
for {
  <-ticker.C
  for _, logger:= range loggers {
    atomic. StoreInt64(&logger.count, 0)
  }
}
```

l2isbad:
@Wing924 i think i fixed all bugs, any comments?

l2isbad:
@Wing924 :heart:

PR #24, opened 2018-08-09
helpers.web refactor

l2isbad:
Changes:

1.
```go
	Header   map[string]string `yaml:"headers"`
	Username string            `yaml:"username"`
	Password string            `yaml:"password"`
```
were moved to `RawRequest` struct.

2.  `Client` defaults:
 * `InsecureSkipVerify` `true` => `false`
 * `FollowRedirect` `false` => `true`


internal/pkg/helpers/web/web.go
```diff
@@ -0,0 +1,7 @@
+package web
+
+// Web is a struct with embedded RawRequest and RawClient.
+type Web struct {
```
Wing924: `Web` name is not so good.
How about `HTTP`?
cf. teregraf use http: https://github.com/influxdata/telegraf/tree/master/plugins/inputs/http

l2isbad:
What about `InsecureSkipVerify` i feel like we don't even need such option

Wing924:
InsecureSkipVerify is very very useful.
Now a days, some apps force you to use https :(
Because many companies don't have in house PKI, they use self-signed certs...

Wing924:
Btw, I think `InsecureSkipVerify` should be `false` by default, but we need to provide a option to enable it.

l2isbad:
Ok, i think this is ready for now. @Wing924

PR #21, opened 2018-08-02
[Super WIP] Refactor job

Wing924:
super WIP, don't merge it.
Fix #19 

- [x] handle signal
- [ ] unit test
- [ ] etc.

Issue #19, opened 2018-07-31
Refactor Plugin/Job


Wing924:
Currently, GoDPlugin and job part are very complex and hard to read.
I'm doing something refactor now.

My goal:

- [ ]  The main should be responsible for args and ENV.
- [ ]  The GoDPlugin should be responsible for controlling jobs and global time tick.
- [ ] The jobs should be responsible for doing real tasks.

l2isbad:
Hm, `GoDPlugin ` is not complex at all, it has only 4 steps - `create`, `init`, `check` and `start` job.  I found it very easy.

Wing924:
Try to write unit tests and you will find how complex they are.

l2isbad:
Wy `global time tick` when every job kinda has their own uppdate freq?

Wing924:
global time ticker is easier to sync time.
we expect all modules run at the time and if a module run longer than 1sec, it shoud skip next run, right?

l2isbad:
> we expect all modules run at the time and if a module run longer than 1sec, it shoud skip next run, right?

No and no.

l2isbad:
> global time ticker is easier to sync time.

There is nothing to sync. Do a run, and calc the next based on spent time.

Wing924:
what will happen if a module run more than 1sec?

Wing924:
https://play.golang.org/p/L4CzszOHxrX

This example shows how easy to use global clock with chan.

l2isbad:
i agree that chan's, tickers is go way, but keep in mind - evey job (not module) has its own update freq, data collections must be aligned to seconds boundaries. I have no clue if std Ticker provide that functionality (it should tick at 00:01, 00:02, not at 00.01:010, 00.02:010)

Wing924:
> data collections must be aligned to seconds boundaries

I'll check it.

l2isbad:
Read [this article](https://github.com/firehol/netdata/wiki/External-Plugins#writing-plugins-properly) carefully . So i basically implement Costa's `The best way to iterate at a constant pace is this pseudo code:`.

l2isbad:
The most important:

1. We must make sure we are always aligned with the netdata daemon. Mandatory.  **Accurately** calculate time for the next run. We have to use monotonic here.
2. We have to **accurately** calculate`time passed since the last run`. Calculate it using real time, not monotonic. Mandatory.

l2isbad:
It's default values per module (per module, not per job).

[...]

PR #25, opened 2018-10-15
[WIP] Big refactor

l2isbad:
let's get it started
Fix :#19
Fix :#21


godplugin/configs.go
```diff
@@ -0,0 +1,109 @@
...
+	MaxProcs   int             `yaml:"max_procs"`
+	Modules    map[string]bool `yaml:"modules"`
+}
+
+func (c *Config) Load(filename string) error {
+	file, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	err = yaml.NewDecoder(file).Decode(c)
```
Wing924:
```suggestion
	return yaml.NewDecoder(file).Decode(c)
```

godplugin/configs.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+func NewConfig() *Config {
+	return &Config{
+		DefaultRun: true,
+		Enabled:    true,
+	}
+}
+
+type Config struct {
```
Wing924:
```suggestion
type GlobalConfig struct {
```

godplugin/configs.go
```diff
@@ -0,0 +1,105 @@
...
+	if run, ok := c.Modules[module]; ok {
+		return run
+	}
+	if explicit {
+		return false
+	}
+	return c.DefaultRun
+}
+
+type rawModConfig []yaml.MapSlice
```
Wing924:
```suggestion
type ModuleConfig struct {
  Global ModuleGlobalConfig `yaml:"global"`
  Jobs []JobConfig `yaml:"jobs"`
}

type ModuleGlobalConfig  struct {...}
type JobConfig struct {...}
```

godplugin/module_config.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+type moduleConfig struct {
+	Global rawConfig   `yaml:"global"`
+	Jobs   []rawConfig `yaml:"jobs"`
+}
+
+func (m *moduleConfig) load(filename string) error {
+	file, err := os.Open(filename)
+	defer file.Close()
```
Wing924: move to below err check

godplugin/plugin.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+var log = logger.New("plugin", "main")
+var validate = validator.New()
+
+func New() *Plugin {
+	return &Plugin{
+		modules:   make(modules.Registry),
+		loopQueue: make([]Job, 0),
+		checkCh:   make(chan Job),
```
Wing924: `make(chan Job, 1)`
at least 1 item

godplugin/plugin.go
```diff
@@ -0,0 +1,261 @@
...
+	Shutdown()
+}
+
+var log = logger.New("plugin", "main")
+var validate = validator.New()
+
+func New() *Plugin {
+	return &Plugin{
+		modules:   make(modules.Registry),
+		loopQueue: make([]Job, 0),
```
Wing924: don't need it it.
loopQueue default init to empty slice

godplugin/plugin.go
```diff
@@ -0,0 +1,261 @@
...
+	p.MainLoop()
+}
+
+func (p *Plugin) MainLoop() {
+	log.Info("start main loop")
+	var clock int
+	tk := ticker.New(time.Second)
+
+	for {
+		select {
```
Wing924: we don't need select case here

godplugin/plugin.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+}
+
+func (p *Plugin) checkJobs() {
+	started := make(map[string]bool)
+
+	for job := range p.checkCh {
+		if started[job.FullName()] {
+			log.Warningf("skipping %s[%s]: already served by another job", job.ModuleName(), job.Name())
```
Wing924: it's normal, not warning. info is ok

godplugin/plugin.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+}
+
+func (p *Plugin) createJobs() []Job {
+	var jobs []Job
+	for name, creator := range p.modules {
+		var modConfig moduleConfig
+
+		// FIXME:
+		err := modConfig.load(fmt.Sprintf("/opt/go.d/%s.conf", name))
```
Wing924: hard code path

godplugin/plugin.go
```diff
@@ -0,0 +1,261 @@
...
+
+		if err != nil {
+			log.Errorf("skipping %s: %s", name, err)
```
Wing924: `"skipping %s: %v"`

godplugin/plugin.go
```diff
@@ -0,0 +1,261 @@
...
+			continue
+		}
+
+		if len(modConfig.Jobs) == 0 {
+			log.Errorf("skipping %s: config 'Jobs' section is empty or not exist", name)
+			continue
+		}
+
+		jobName := func(conf rawConfig) interface{} {
```
Wing924: why create a func in loop?

godplugin/module_config.go
l2isbad: reason?

godplugin/plugin.go
l2isbad: it's under FIXME, i know it

godplugin/plugin.go
l2isbad: actually for no reason lol

godplugin/module_config.go
Wing924: if err is not nil, file may be nil

l2isbad: `Close` can handle it, since we are not checking `Close` result it's OK to leave it here

l2isbad:
This is huge.

PR #34, opened 2018-11-21
remove panicked job from the plugin loop queue

l2isbad:
Fixes: #27

### DIFF
--- a/godplugin/plugin.go
+++ b/godplugin/plugin.go
@@ -48,9 +48,6 @@ func New() *Plugin {
 	return &Plugin{
 		modules: make(modules.Registry),
 		checkCh: make(chan Job, 1),
-		loopQueue: jobQueue{
-			mux: sync.Mutex{},
-		},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


godplugin/plugin.go
```diff
@@ -47,6 +48,43 @@ func New() *Plugin {
...
+}
+
+func (q *jobQueue) add(job Job) {
+	q.mux.Lock()
+	defer q.mux.Unlock()
+
+	q.queue = append(q.queue, job)
+}
+
+func (q *jobQueue) pop(fullName string) Job {
```
Wing924: pop is not a good name because it means remove last item and return that

godplugin/plugin.go
```diff
@@ -47,6 +48,43 @@ func New() *Plugin {
...
 		modules: make(modules.Registry),
 		checkCh: make(chan Job, 1),
+		loopQueue: &jobQueue{
+			mux: &sync.Mutex{},
+		},
+	}
+}
+
+type jobQueue struct {
+	mux   *sync.Mutex
```
Wing924: why use ptr?

godplugin/plugin.go
l2isbad: agreed

Issue #38, opened 2018-11-22
refactor web_log


ilyam8:

l2isbad:
Fixes: https://github.com/netdata/netdata/issues/3111

PR #74, opened 2018-12-10
web_log fix p1

ilyam8:


ilyam8:
Ok, a lot of todos, but this version is much better, it uses https://github.com/hpcloud/tail and csv parser. 

I will merge and open a new PR with todos and fixes

ilyam8:
Fixes : #38

PR #90, opened 2018-12-22
web client refactor

ilyam8:
* move `ProxyName` to web.Request
* move `ProxyPassword` to web.Request
* remove web.HTTPClient (new func now returns *http.Client)

### DIFF
--- a/pkg/web/request_test.go
+++ b/pkg/web/request_test.go
@@ -1,36 +1,62 @@
 package web
 
 import (
+	"encoding/base64"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
 var (
-	username    = "user"
-	password    = "password"
-	headerKey   = "X-Api-Key"
-	headerValue = "secret"
+	username      = "user"
+	password      = "password"
+	headerKey     = "X-Api-Key"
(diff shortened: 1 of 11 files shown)


pkg/web/request_test.go
```diff
@@ -1,36 +1,62 @@
...
+	assert.IsType(t, (*http.Request)(nil), req)
 
+	user, pass, ok := req.BasicAuth()
 	assert.True(t, ok)
 	assert.True(t, user == username && pass == password)
-	assert.True(t, httpReq.Header.Get(headerKey) == headerValue)
+
+	user, pass, ok = parseBasicAuth(req.Header.Get("Proxy-Authorization"))
+	assert.True(t, ok)
+	assert.True(t, user == proxyUsername && pass == proxyPassword)
```
Wing924: should be:
```
assert.Equal(t, user, proxyUsername)
assert.Equal(t, pass, proxyPassword)
```

pkg/web/request_test.go
```diff
@@ -1,36 +1,62 @@
...
+
+	assert.True(t, req.Header.Get(headerKey) == headerValue)
```
Wing924: assert.Equal(t, headerValue, req.Header.Get(headerKey))

PR #102, opened 2019-01-03
Add OracleDB module

kataras:
1. **Processes**
  * total

2. **Sessions**
  * total
  * active
  * inactive

2. **Activity**
 * parse count (total)
 * execute count
 * user commits
 * user rollbacks

3. **Wait Times**
 * configuration
 * administrative
 * system I/O
 * application
 * concurrency
 * commit
 * network
 * user I/O
 * other

4. **Tablespace Size**
 * max_bytes_system
 * free_bytes_system
 * bytes_system
 * max_bytes_sysaux
 * free_bytes_sysaux
 * bytes_sysaux
 * max_bytes_users
 * free_bytes_users
 * bytes_users
 * max_bytes_temp
 * free_bytes_temp
 * bytes_temp

### configuration

```yaml
jobs:
  - name: local
    dsn: SYSTEM/Oracle12345@ORCL

```


go.mod
```diff
@@ -21,11 +21,10 @@ require (
...
 	golang.org/x/crypto v0.0.0-20181203042331-505ab145d0a9 // indirect
 	golang.org/x/net v0.0.0-20181011144130-49bb7cea24b1 // indirect
-	golang.org/x/sync v0.0.0-20180314180146-1d60e4601c6f // indirect
 	golang.org/x/sys v0.0.0-20181011152604-fa43e7bc11ba // indirect
 	gopkg.in/fsnotify.v1 v1.4.7 // indirect
 	gopkg.in/go-playground/assert.v1 v1.2.1 // indirect
+	gopkg.in/goracle.v2 v2.11.0 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 	gopkg.in/yaml.v2 v2.2.1
-	layeh.com/radius v0.0.0-20181224030715-9a016ab9b9ec
```
ilyam8: This is no good.
And i think better to fork the repo and make a PR from fork and not directly to master.

modules/all/all.go
```diff
@@ -15,4 +15,5 @@ import (
 	_ "github.com/netdata/go.d.plugin/modules/solr"
 	_ "github.com/netdata/go.d.plugin/modules/springboot2"
 	_ "github.com/netdata/go.d.plugin/modules/weblog"
+	// Oracle DB should be configured with specific dsn.
```
ilyam8: Should be imported in `all` package. It adds module to the registry.

modules/oracledb/README.md
```diff
@@ -0,0 +1,79 @@
+# rabbitmq
```
ilyam8: not :rabbit: mq

modules/oracledb/charts.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+	},
+	{
+		ID:    "sessions",
+		Title: "Total Sessions",
+		Ctx:   "oracledb.sessions",
+		Dims: Dims{
+			{ID: "sessions_total", Name: "sessions_total"},
+			{ID: "sessions_active", Name: "sessions_active"},
+			{ID: "sessions_inactive", Name: "sessions_inactive"},
```
ilyam8: i don't like that one chart contains total and specific types of connections.

My suggestion (or):
 * move `total` to a separate chart
 * remove total, change chart algorithm to `stacked` (total will be `active` + `inactive`)

And, this is true for **all charts**:
 1. don't copy dimension id to `name`

name
> is the name that will be presented to the user instead of id in type.id. This means that only the id part of type.id is changed. When a name has been given, the chart is index (and can be referred) as both type.id and type.name. You can set name to '', or null, or (null) to disable it.

Should be:

```go
			{ID: "sessions_active", Name: "active"},
			{ID: "sessions_inactive", Name: "inactive"}
```

 2. please add `Units` to all charts.

```go
		ID:    "sessions",
		Title: "Sessions",
		Ctx:   "oracledb.sessions",
         	Units:   "sessions",
		Dims: Dims{
			{ID: "sessions_total", Name: "total"},
			{ID: "sessions_active", Name: "active"},
			{ID: "sessions_inactive", Name: "sessions_inactive"},
```

[Dimension algorithm](https://docs.netdata.cloud/collectors/plugins.d/#dimension) by default is `absolute`. If you need `incremental` you have to set it explicitly

modules/oracledb/oracledb.go
```diff
@@ -0,0 +1,629 @@
...
+	"database/sql"
+	"strconv"
+	"strings"
+
+	"github.com/netdata/go.d.plugin/modules"
+
+	"gopkg.in/goracle.v2"
+)
+
+const defaultDSN = "SYSTEM/Oracle12345@ORCL" // there is no such thing as default connstr/dsn, but...
```
ilyam8: If there is no such thing as default dsn please don't set default. Just check in `Init` if it's set and if not return `false`.

modules/oracledb/oracledb.go
```diff
@@ -0,0 +1,629 @@
...
+		Create: func() modules.Module { return New(defaultDSN) },
+	}
+
+	modules.Register("oracledb", creator)
+}
+
+// OracleDB oracledb module.
+type OracleDB struct {
+	modules.Base
+	DSN string `json:"dsn,omitempty" yaml:"dsn,omitempty"`
```
ilyam8: Why do we need `json`? We use `yaml` for config files.

modules/oracledb/oracledb.go
```diff
@@ -0,0 +1,629 @@
...
+// Charts creates Charts.
+func (m *OracleDB) Charts() *Charts {
+	return charts.Copy()
+}
+
+// Collect collects health checks and metrics for OracleDB.
+func (m *OracleDB) Collect() map[string]int64 {
+	metrics := make(map[string]int64)
+
+	db, err := sql.Open("goracle", m.DSN)
```
ilyam8: I see you reopen connection on every iteration. How about to reuse same connection and reopen it (reconnect) only in case of failures?

modules/oracledb/oracledb.go
```diff
@@ -0,0 +1,629 @@
...
+	if err != nil {
+		m.Errorf("error on opening a connection with the oracle database [%s]: %v", m.DSN, err)
+		return metrics
+	}
+	defer db.Close()
+
+	err = m.collectProcesses(db, metrics)
+	if err != nil {
+		m.Errorf("error on collecting processes: %v", err)
```
ilyam8: I think this is depends on error. If connection is broken there is no reason to continue.

Example: network connectivity failes, oracle service restart. We need to return nil immediately and try to reconnect on next iteration.

modules/oracledb/oracledb.go
```diff
@@ -0,0 +1,629 @@
...
+
+	metrics["total_sessions"] = totalCount
+	metrics["active_sessions"] = activeCount
+	metrics["inactive_sessions"] += inactiveCount
+
+	return nil
+}
+
+func cleanActivityName(s string) string {
+	return strings.Replace(strings.Replace(strings.Replace(s, "(", "", -1), ")", "", -1), " ", "_", -1)
```
ilyam8: this is hard to read, how about to use `string.Replacer` ?

modules/oracledb/oracledb.go
kataras: Yes, opening a new connection on every itaration was not accidental, other metric systems do the same thing but if you want to re-use the connection I'll do it now, no problem.

ilyam8: It's not something i want, if reusing same connection is more effective in terms of performance we should do it. If it is not open/close is ok too.

I think we don't need `ping` and it is ok to assume that connection is alive. If `db.query` returns an error (maybe some type of error, not an expert) the flow should be:

modules/oracledb/oracledb.go
kataras: ✅

modules/oracledb/oracledb.go
kataras: ✅

modules/oracledb/oracledb.go
kataras: ✅

modules/oracledb/oracledb.go
kataras: ✅

modules/oracledb/README.md
kataras: ✅

modules/all/all.go
kataras: ✅

go.mod
kataras: ✅

modules/oracledb/oracledb.go
```diff
@@ -0,0 +1,657 @@
...
+	return true
+}
+
+func (m *OracleDB) openConnection() error {
+	if m.db != nil {
+		if err := m.db.Ping(); err != nil {
+			m.db.Close()
+			m.db = nil
+
+			return m.openConnection()
```
ilyam8: Just log the error and return nil. It will try to reconnect on every iteration, this is ok flow.

ilyam8: and return an error ofc

ilyam8: oh, i am wrong, your code is ok
sorry

modules/oracledb/charts.go
```diff
@@ -0,0 +1,108 @@
...
+			{ID: "system_service_response_time", Name: "system_service_response_time"},
+			{ID: "system_user_rollbacks", Name: "system_user_rollbacks"},
+			{ID: "system_sorts_per_user_call", Name: "system_sorts_per_user_call"},
+			{ID: "system_rows_per_sort", Name: "system_rows_per_sort"},
+			{ID: "system_disk_sorts", Name: "system_disk_sorts"},
+			{ID: "system_memory_sorts_ratio", Name: "system_memory_sorts_ratio"},
+			{ID: "system_database_wait_time_ratio", Name: "system_database_wait_time_ratio"},
+			{ID: "system_session_limit_usage", Name: "system_session_limit_usage"},
+			{ID: "system_session_count", Name: "system_session_count"},
+			{ID: "system_temp_space_used", Name: "system_temp_space_used"},
```
ilyam8: should be splitted into several charts

important: all dimensions must have same `Units`

modules/oracledb/charts.go
kataras: Same for `processes`? the dim should be: `Dim {  ID: "processes", Name: "processes", Algo: modules.Incremental }` for example?

kataras: The dim should be: `{  ID: "sessions_active", Name: "active", Algo: modules.Incremental }` for example? We should start writing some documentation for these things for the go.d.plugin...

modules/oracledb/oracledb.go
```diff
@@ -0,0 +1,657 @@
...
+		Create: func() modules.Module { return New("") },
+	}
+
+	modules.Register("oracledb", creator)
+}
+
+// OracleDB oracledb module.
+type OracleDB struct {
+	modules.Base
+	DSN string `yaml:"dsn,omitempty"`
```
ilyam8: `omitempty` is not needed

> The "omitempty" option specifies that the field should be omitted from the **encoding** if

modules/oracledb/oracledb.go
```diff
@@ -0,0 +1,657 @@
...
+	db  *sql.DB
+}
+
+// New creates OracleDB mod.
+func New(dsn string) *OracleDB {
```
ilyam8: we don't need input params for the New, please remove

modules/oracledb/oracledb.go
```diff
@@ -0,0 +1,657 @@
...
+	err := m.db.Close()
+	if err != nil {
+		m.Errorf("cleanup: error on closing the oracle database [%s]: %v", m.DSN, err)
+	}
+}
+
+// Init makes initialization of the OracleDB mod.
+func (m *OracleDB) Init() bool {
+	if m.DSN == "" {
+		return false
```
ilyam8: please add error message

modules/oracledb/charts.go
ilyam8: it's netdata api, it's[ documented](https://docs.netdata.cloud/collectors/plugins.d/#external-plugins-api)

ilyam8:
from @Wing924 

Merging it we breaking go's single binary rule. Binary will need some shared libs to run, which may not exist on target system.

For example it may not run on alpine linux.

@kataras @paulfantom

ilyam8:
Just thought: as a workaround we coould add `oracledb` as a standalone plugin later.

PR #103, opened 2019-01-03
consul

ilyam8:


modules/consul/consul.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// Consul consul module
+type Consul struct {
+	modules.Base
+
+	web.HTTP `yaml:",inline"`
+
+	MaxChecks int `yaml:"max_checks"`
+	Token     string
```
Wing924: I think `ACLToken` is better name

PR #104, opened 2019-01-08
MySQL Module

kataras:
There are more but need to be tested before push, stay tuned.

Edit 16 Jan 2018:

1) Slave stats are not retrieved by my machine, will find examples online, we need only three cols.
2) README.md and charts.go are TODO
3) main implementation is done.


.vscode/settings.json
```diff
@@ -0,0 +1,63 @@
...
+    "git.enabled": true,
+    "extensions.ignoreRecommendations": true,
+    "git.enableSmartCommit": true, // I always turn off git on my editors, I am managing git diffs and pushes by programs like Github Desktop or Kraken.
+    "search.exclude": {
+        "**/node_modules": true,
+        "**/bower_components": true,
+        "vendor/**": true
+    },
+    "editor.tabCompletion":"on"
+}
```
ilyam8: i think we don't need this

ilyam8:
@kataras i changed `globalStats` to `map[sting]bool` because it is less error prone (no way to add duplicate value) and faster to check if value in `globalStats`

ilyam8:
@kataras that is what i got for user without permissions to query replication status

> [ ERROR ] mysql[remote] logger.go:99 error on collecting slave status: Error 1227: Access denied; you need (at least one of) the SUPER, REPLICATION CLIENT privilege(s) for this operation

and job died

replication client permissions shouldn;t be mandatory for module to work

modules/mysql/mysql.go
```diff
@@ -0,0 +1,524 @@
...
+		}
+	}
+
+	// test the connectivity here.
+	if err := m.openConnection(); err != nil {
+		return false
+	}
+
+	if min, got := CompatibleMinimumVersion, m.getMySQLVersion(); min > 0 && got < min {
+		m.Warningf("running with uncompatible mysql version [%v<%v]", got, min)
```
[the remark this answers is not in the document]
kataras: No, the idea was that, a metrics collector which has a `Version()` method would run through a loop inside the main module, compare `getCurrentMySQLVersion()` version vs collector's `Version()` which should be >= in order to that collect to run, otherwise a warning would show up.

config/go.d/mysql.conf
```diff
@@ -0,0 +1,87 @@
...
+# [ JOBS ]
+jobs:
+  - name: local
+    dsn: netdata@tcp(127.0.0.1:3306)/
+  - name: local2
+    dsn: root@/
+  - name: socket
+    dsn: root@unix(/var/run/mysqld/mysqld.sock)/
+  - name: socket2
+    dsn: root@unix(/usr/local/var/mysql/mysql.sock)/
```
ilyam8: > #  - Jobs with the same name are mutually exclusive. Only one of them will be allowed running at any time.

We need only one of them be running at a time. Please change name to `local` to all of them.

ilyam8: This is the feature - you add several jobs with same name but different parameters and the first that succed will run, all other will be dropped.

https://github.com/netdata/go.d.plugin/blob/140abe8d3ade403813619abc93a193f88e797ad3/godplugin/jobstart.go#L12-L16

modules/mysql/mysql.go
```diff
@@ -0,0 +1,575 @@
...
+	"Slave_IO_Running":      slaveRunning,
+}
+
+// hasReplPriv returns if the current user has privileges for the "SHOW SLAVE STATUS" query.
+func (m *MySQL) hasReplPriv() bool {
+	if m.user == "" {
+		return false
+	}
+
+	rows, err := m.db.Query("select User,Super_priv,Repl_client_priv from mysql.user")
```
ilyam8: this doesn't work, user has no permissions to execute SELECT queries

netdata suggested

```
create user 'netdata'@'localhost';
grant usage on *.* to 'netdata'@'localhost';
grant replication client on *.* to 'netdata'@'localhost';  # optional
flush privileges;

File: modules/mysql/mysql.go [hunk 5]
```diff
@@ -0,0 +1,575 @@
...
+
+		// fmt.Printf("var name: %s | value: %s\n", varName, value)
+
+		if valFn, ok := slaveStats[varName]; ok { // hmmm...
+			value := valFn(value)
+			metrics[toKey(varName)] = value
+		}
+	}
+
+	return nil
```
ilyam8: Doesn't work

i found https://github.com/gdaws/mysql-slave-status

> [ ERROR ] mysql[goland] logger.go:99 error on collecting slave status: sql: expected 47 destination arguments in Scan, not 2

ilyam8: i did a quick check (replaced `collectSlaveStatus` code with code from `mysql-slave-status`)

it, works

`
map[Relay_Log_Pos:{1183 true} Replicate_Wild_Do_Table:{ true} Last_Errno:{0 true} Exec_Master_Log_Pos:{895 true} Until_Log_File:{ true} Last_IO_Errno:{0 true} Last_IO_Error:{ true} Master_SSL_Cipher:{ true} Replicate_Ignore_Server_Ids:{ true} Master_Server_Id:{1 true} Relay_Master_Log_File:{mysql-bin.000002 true} Replicate_Do_Table:{ true} Master_Log_File:{mysql-bin.000002 true} Master_SSL_CA_File:{ true} Master_SSL_Crl:{ true} Replicate_Do_Domain_Ids:{ true} Parallel_Mode:{conservative true} Master_SSL_Crlpath:{ true} Connect_Retry:{60 true} Replicate_Ignore_DB:{ true} Skip_Counter:{0 true} Master_SSL_Key:{ true} Seconds_Behind_Master:{0 true} Last_SQL_Errno:{0 true} Last_SQL_Error:{ true} Using_Gtid:{No true} Slave_IO_State:{Waiting for master to send event true} Master_User:{slave_user true} Replicate_Do_DB:{ true} Until_Log_Pos:{0 true} Gtid_IO_Pos:{ true} Replicate_Ignore_Domain_Ids:{ true} Master_SSL_Cert:{ true} Master_Host:{100.127.0.91 true} Read_Master_Log_Pos:{895 true} Replicate_Wild_Ignore_Table:{ true} Last_Error:{ true} Relay_Log_Space:{1833 true} Master_SSL_Allowed:{No true} Master_SSL_CA_Path:{ true} Master_Port:{3306 true} Relay_Log_File:{mysql-relay-bin.000003 true} Slave_IO_Running:{Yes true} Slave_SQL_Running:{Yes true} Replicate_Ignore_Table:{ true} Until_Condition:{None true} Master_SSL_Verify_Server_Cert:{No true}]
`

ilyam8:
github.com/go-sql-driver/mysql

> Automatic handling of broken connections

It works :smile:

PR #117, opened 2019-01-15
add limit to cachedMatcher

ilyam8:


pkg/matcher/cache.go
```diff
@@ -1,31 +1,83 @@
...
+		} else {
+			cm.cache = newLRUCache(limit)
 		}
+		return cm
 	}
 }
 
 func (m *cachedMatcher) Match(b []byte) bool {
 	s := string(b)
+	if m.cache == nil {
```
Wing924: it never be true because `m.cache` must not be `nil`

pkg/matcher/cache.go
```diff
@@ -35,6 +87,10 @@ func (m *cachedMatcher) Match(b []byte) bool {
 }
 
 func (m *cachedMatcher) MatchString(s string) bool {
+	if m.cache == nil {
```
Wing924: it never be true because m.cache must not be nil

pkg/matcher/cache.go
```diff
@@ -1,31 +1,86 @@
...
-import "sync"
+import (
+	"sync"
+
+	"github.com/hashicorp/golang-lru"
+)
+
+type cache interface {
+	Get(key string) (result bool, exist bool)
+	Len() int
```
Wing924: do we need `Len()`?

modules/activemq/activemq.go
```diff
@@ -29,10 +29,11 @@ var (
 )
 
 var (
-	defMaxQueues   = 50
-	defMaxTopics   = 50
-	defURL         = "http://127.0.0.1:8161"
-	defHTTPTimeout = time.Second
+	defMaxQueues       = 50
```
Wing924: `def` usually means not `default` but `define`.
I think use `default` is more readable.

ilyam8:
thx @Wing924

Issue #130, opened 2019-01-21
web_log should use correct metrics type


Wing924:
resp_time is not counter, it should use gauge.

ref. netdata/netdata#3907

ilyam8:
found
https://github.com/netdata/netdata/pull/1758#issuecomment-279150948

> I think the web_log plugin has an major issue. You are passing all metrics as absolute.

Wing924:
@ilyam8 
I'm talking about golang's version
https://github.com/netdata/go.d.plugin/blob/master/modules/weblog/charts.go#L80-L82

ilyam8:
Same. Netdata does interpolation for `incremental` values and shows `absolute` as-is, so incremental i always preferable

@ktsaou

Wing924:
@ilyam8 

> Same. Netdata does interpolation for `incremental` values and shows `absolute` as-is, so incremental i always preferable

It's not true. You can check netdata/netdata#3907 and found what's wrong.

Issue #88, opened 2018-12-20
web_log extensive testing/optimization


ilyam8:
hpcloud/tail consumes a lot of cpu, shouldn't be used

Wing924:
hpcloud/tail  looks slow, we need fsnotify.
cf. https://github.com/grafana/loki/blob/68e1a049a0/pkg/promtail/targets/filetarget.go

Wing924:
not easy as it looks

ilyam8:
i'd say it doesn't look easy.

PR #141, opened 2019-01-25
reafctor web_log

Wing924:
Fixes: #88
Fixes: #130
Fixes: #286

Fixes: netdata/netdata#7245


cakrit:
Any updates here? It seems to be blocking https://github.com/netdata/netdata/pull/4900

Wing924:
@cakrit I'll finish it soon.
btw, logparser module looks much different from web_log module because web_log only handle access log while logparser handle generic logs.

PR #146, opened 2019-02-06
Change "temporary disabled" to WorkInProgress

cakrit:
The title on the readmes shows up in netlify under 3 levels, so it needs to be short.


ilyam8:
changed title because of the WIP check :smile:

PR #190, opened 2019-04-08
support springboot2 URI filter

### LINKED ISSUE #189: springboot2 feature: add uri filters
Right now the springboot2 modules which depends on prometheus metrics doesn't take any parameters.  I suggest a yaml parameter be added that can be a list of regexp patterns to exclude uri's from the metrics responses (so that actuator, etc.) can be filtered.  Maybe the parm would be called filter_uri_regexp (which is a yaml list of regexps)?

Bonus if you enable both a filter (remove) and a set (only do for specified uri's).

I'm tempted to fork and attempt, but I'm not a "go" expert... so not sure if I'd do things in the most efficient manner.

The reason for this feature, right now the response counts count all bad, unknown, and actuator endpoints... and you might be interested only in the counts for your custom endpoints.

Wing924:
fix #189

### DIFF
--- a/pkg/matcher/expr.go
+++ b/pkg/matcher/expr.go
@@ -1,0 +1,78 @@
+package matcher
+
+import "fmt"
+
+type (
+	// SimpleExpr is a simple expression to describe the condition:
+	//     (includes[0].Match(v) || includes[1].Match(v) || ...) && !(excludes[0].Match(v) || excludes[1].Match(v) || ...)
+	SimpleExpr struct {
+		Includes []string `yaml:"includes" json:"includes"`
+		Excludes []string `yaml:"excludes" json:"excludes"`
+
+		includes []Matcher
+		excludes []Matcher
+	}
+)
+
+// Parse parse the given matchers in Includes and Excludes
+func (s *SimpleExpr) Parse() error {
+	if len(s.Includes) > 0 {
+		s.includes = make([]Matcher, 0, len(s.Includes))
+		for _, item := range s.Includes {
+			m, err := Parse(item)
+			if err != nil {
+				return fmt.Errorf("parse matcher %q error: %v", item, err)
+			}
+			s.includes = append(s.includes, m)
+		}
+	}
+	if len(s.Excludes) > 0 {
+		s.excludes = make([]Matcher, 0, len(s.Excludes))
+		for _, item := range s.Excludes {
+			m, err := Parse(item)
+			if err != nil {
+				return fmt.Errorf("parse matcher %q error: %v", item, err)
+			}
+			s.excludes = append(s.excludes, m)
+		}
+	}
+	return nil
+}
+
+// Match match against []byte
+func (s *SimpleExpr) Match(b []byte) bool {
+	if len(s.includes) > 0 {
+		for _, m := range s.includes {
+			if !m.Match(b) {
+				return false
+			}
+		}
+	}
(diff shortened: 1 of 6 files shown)


ilyam8:
@Wing924 about the SimpleExpr thing, i think we need use logical, since we have it

https://github.com/netdata/go.d.plugin/blob/master/modules/coredns/filter.go

ilyam8:
So i suggest 2 things:
 - simplify SimpleExpr using logical funcs
 - return true if Include/Exclude matcher matches, but use Not for Exclude (https://github.com/netdata/go.d.plugin/blob/a212685749537aa85b294842d90f9f26b1faaee0/modules/coredns/collect.go#L284-L285)

pkg/matcher/expr.go
```diff
@@ -0,0 +1,52 @@
...
+		m, err := Parse(item)
+		if err != nil {
+			return fmt.Errorf("parse matcher %q error: %v", item, err)
+		}
+		excludes = Or(excludes, m)
+	}
+
+	if len(s.Includes) == 0 {
+		includes = TRUE()
+	}
```
ilyam8: if `Includes` and `Excludes` are empty matcher should return `false`, not `true`

pkg/matcher/expr.go
```diff
@@ -0,0 +1,52 @@
+package matcher
+
+import "fmt"
+
+type (
+	// SimpleExpr is a simple expression to describe the condition:
+	//     (includes[0].Match(v) || includes[1].Match(v) || ...) && !(excludes[0].Match(v) || excludes[1].Match(v) || ...)
+	SimpleExpr struct {
+		Includes []string `yaml:"includes" json:"includes"`
+		Excludes []string `yaml:"excludes" json:"excludes"`
```
ilyam8: i suggest to change to `Include` and `Exlcude`

Wing924: I use includes and excludes because test kitchen use this syntax:
<https://docs.chef.io/kitchen.html#syntax>

pkg/matcher/expr.go
Wing924: if Includes is empty, we should default them to `true` or empty filter config will lead to exclude everything.

ilyam8: Empty config == don't use matcher.

Otherwise we need to add bool param to every filer.

https://github.com/netdata/go.d.plugin/blob/a212685749537aa85b294842d90f9f26b1faaee0/modules/coredns/collect.go#L48-L53

like `if cd.DoPerServerStats { ... }`

Wing924: for example:
```
filter:
  excludes:
  - = foo
```
only `foo` should be excluded, others should be pass.

if you set empty `includes` false, everything will be excluded.

ilyam8: or add `IsEmpty` or something method to it, to know when to use it

pkg/matcher/expr.go
ilyam8: ok, maybe my eglish is bad, this `s doesn't make sense to me in that context

pkg/matcher/expr.go
Wing924: `if filter == TRUE() {...}` is not enough?

ilyam8: Enough, but please make default (empty)`FALSE`, because it is logical - no matchers no match

PR #210, opened 2019-04-30
scaleio collector

ilyam8:
Fixes: netdata/netdata#4825


ilyam8:
This is only initial implementation

PR #233, opened 2019-06-19
fix status_accepted due to the latest config file

tru2dagame:
https://github.com/netdata/go.d.plugin/blob/master/config/go.d/httpcheck.conf#L174


ilyam8:
Thanks @tru2dagame !

Issue #230, opened 2019-06-13
dnsmasq_dhcp improvements


ilyam8:
1. Folowwing parameters should be automatically detected during startup.
 >  The DHCP server needs somewhere on disk to keep its lease database.
 This defaults to a sane location, but if you want to change it, use
 the line below.

`dhcp-leasefile=/var/lib/misc/dnsmasq.leases`

 >  Include another lot of configuration options.
conf-file=/etc/dnsmasq.more.conf

`conf-dir=/etc/dnsmasq.d`

> Include all files in a directory which end in .conf

`conf-dir=/etc/dnsmasq.d/,*.conf`

___

2. Dnsmasq doesn't add static IPs to the leases database. To take them into account we need to read dnsmasq configuration files.

PR #239, opened 2019-07-03
modules/dnsmasq_dhcp: respect conf-file/conf-dir 

kovetskiy:
#230 

dnsmasq_dhcp respects conf-file/conf-dir directive

* dnsmasq_dhcp respects conf-file=<path> configuration lines and reads
    specified files
* dnsmasq_dhcp respects conf-dir=<path>[,<suffix-include-or-exclude>...]
    configuration lines and recursively reads specified directories and
    files in these directories
* fix unexpected behavior: dnsmasq_dhcp reads contents of 'dnsmasq.d' directory
    by default (dnsmasq doesn't do that if 'conf-dir' is not specified)

Haven't added the following option:
```
The DHCP server needs somewhere on disk to keep its lease database.
This defaults to a sane location, but if you want to change it, use
the line below.
dhcp-leasefile=/var/lib/misc/dnsmasq.leases
```

It will be done in a separated PR.


kovetskiy:
@ilyam8: could you take a look at the PR?

modules/dnsmasq_dhcp/autodetection.go
```diff
@@ -6,28 +6,42 @@ import (
...
+//
+// # Include all the files in a directory except those ending in .bak
+// #conf-dir=/etc/dnsmasq.d,.bak
+//
+// # Include all files in a directory which end in .conf
+// #conf-dir=/etc/dnsmasq.d/,*.conf
+type configDir struct {
+	path          string
+	includeSuffix []string
+	excludeSuffix []string
```
ilyam8: > 	includeSuffix []string
> 	excludeSuffix []string

Why slice?

> conf-dir=/etc/dnsmasq.d/,*.conf

is `conf-dir=/etc/dnsmasq.d/,*.conf, *.cconf` a valid syntax?

modules/dnsmasq_dhcp/autodetection.go
```diff
@@ -46,18 +60,128 @@ func (d *DnsmasqDHCP) findDHCPRanges() ([]string, error) {
...
+			}
+
+			includeDirs = append(includeDirs, dir)
+
+			continue
+		}
+	}
+
+	for _, dir := range includeDirs {
+		dirFiles, err := readConfDir(dir.path, dir.includeSuffix, dir.excludeSuffix)
```
ilyam8: why not `readConfDir(dir)`

i see 

```go
type configDir struct
```

and func name `readConfDir`, i think it is ok to pass `configDir` struct there

modules/dnsmasq_dhcp/autodetection.go
```diff
@@ -46,18 +60,128 @@ func (d *DnsmasqDHCP) findDHCPRanges() ([]string, error) {
...
+		if err != nil {
+			d.Warningf("error during configuration dir %q scanning: %v", dir.path, err)
+		}
+
+		includeFiles = append(includeFiles, dirFiles...)
+	}
+
+	for _, file := range includeFiles[:] {
+		files := d.findConfigs(file)
```
ilyam8: what if `dnsmasq.conf` contains

> conf-file=/etc/dnsmasq.conf

?

why do we need `[:]` ?

modules/dnsmasq_dhcp/autodetection.go
```diff
@@ -128,3 +252,24 @@ func parseDHCPRangeLine(s string) (r string) {
 
 	return fmt.Sprintf("%s-%s", start, end)
 }
+
+func unique(slice []string) []string {
+	result := []string{}
+
+	for _, item := range slice {
+		if !contains(result[:], item) {
```
ilyam8: why do we need `[:]`?

modules/dnsmasq_dhcp/autodetection.go
kovetskiy: Yes, it's valid syntax. I have not found it in dnsmasq's documentation, but found it in dnsmasq's source code, here is the link to dnsmasq's github mirror:

https://github.com/imp/dnsmasq/blob/4e7694d7107d2299f4aaededf8917fceb5dfb924/src/option.c#L1560

I've also checked it practically and it works exactly like that.

modules/dnsmasq_dhcp/autodetection.go
kovetskiy: That's good point about passing structure, I thought about it and decided to abstract the function of struct to show that the func depends only on specific input fields and not on whole struct. But since struct configDir is not going to be changed heavily, we can pass it entirely.

Agreed, the func name is not the best here because it not only reads, but also filters, thought about findConfigsInDir (don't like that `In` in name), scanConfigDir (but scan sounds ambiguous), readAndFilterConfigDir (too long, but sounds idiomatic for Go, like http.ListenAndServe). What do you think?

modules/dnsmasq_dhcp/autodetection.go
kovetskiy: if config will have a recursion, we will run into limit of open files and os.Open() will fail eventually with error `too many open files`, but it's the same case as with dnsmasq itself.

We need `[:]` in order to avoid cycle on new arrived files found by d.findConfigs(), so we re-slice includeFiles and when we add new elements in `includeFiles` they will not appear in re-sliced slice. We could create a new slice and then merge them if you insist on avoiding that trick.

modules/dnsmasq_dhcp/autodetection.go
kovetskiy: good catch, will drop that

modules/dnsmasq_dhcp/autodetection.go
ilyam8: > so we re-slice includeFiles and when we add new elements in includeFiles they will not appear in re-sliced slice

i think `range` already does it and `includeFiles` will point to another slice right after first append. Have you tested it?

ilyam8: i may be wrong there ^^ :thinking: 	not sure

modules/dnsmasq_dhcp/autodetection.go
ilyam8: > readAndFilter

i think we can name it `find` and make it a method of configDir

> findConfigFiles

what do you think?

modules/dnsmasq_dhcp/autodetection.go
ilyam8:
```go
	vv := []int{1, 2, 3}

	println(vv)
	for _, v := range vv {
		vv = append(vv, v * 10)
		println(vv)
	}

	fmt.Println(vv)
```

```
[1 2 3 10 20 30]
[3/3]0xc000018540
[4/6]0xc00001a150  <- different slice
[5/6]0xc00001a150
[6/6]0xc00001a150
```

ilyam8: ok, i think this is a bad example :sweat_smile: 

my point is  re-slicing is not needed in `range`

ilyam8: ok, that is the internals of `for` loop

```
//   for_temp := range
//   len_temp := len(for_temp)
//   for index_temp = 0; index_temp < len_temp; index_temp++ {
//           value_temp = for_temp[index_temp]
//           index = index_temp
//           value = value_temp
//           original body
//   }
```

so, yes, we don't need re-slicing

modules/dnsmasq_dhcp/autodetection.go
kovetskiy: :+1:

kovetskiy:
I've force-pushed the changes.

modules/dnsmasq_dhcp/autodetection.go
```diff
@@ -46,18 +60,128 @@ func (d *DnsmasqDHCP) findDHCPRanges() ([]string, error) {
...
-		if !fi.Mode().IsRegular() || !strings.HasSuffix(fi.Name(), ".conf") {
+		if !fi.Mode().IsRegular() {
 			continue
 		}
-		configs = append(configs, path.Join(confDir, fi.Name()))
+
+		for _, suffix := range dir.excludeSuffix {
+			if strings.HasSuffix(fi.Name(), suffix) {
+				continue FILES
+			}
```
ilyam8: // # Include all the files in a directory except those ending in .bak
// #conf-dir=/etc/dnsmasq.d,.bak

This is a little bit more complex then exclude .bak, include .conf

ilyam8: and i think `findConfigFiles` is a better name, because function finds config files. Otherwise it is not clear what it finds.

ilyam8: i suggest to test how it works when both `exclude` and `include` are set, just to be sure

kovetskiy: check the last commit

kovetskiy:
upd: pushed even more _bad_ file names

modules/dnsmasq_dhcp/autodetection.go
```diff
@@ -128,3 +217,85 @@ func parseDHCPRangeLine(s string) (r string) {
...
+		//     (ent->d_name[0] == '#' && ent->d_name[len - 1] == '#') ||
+		//     ent->d_name[0] == '.')
+		//    continue;
+
+		name := fi.Name()
+		if strings.HasSuffix(name, "~") ||
+			(strings.HasPrefix(name, "#") && strings.HasSuffix(name, "#")) ||
+			strings.HasPrefix(name, ".") {
+			continue
+		}
```
ilyam8: i think this logic deserves its own func `isBadFileName` or something

modules/dnsmasq_dhcp/autodetection.go
```diff
@@ -128,3 +217,85 @@ func parseDHCPRangeLine(s string) (r string) {
...
+			}
+		}
+
+		excluding := false
+		for _, suffix := range dir.excludeSuffix {
+			if strings.HasSuffix(fi.Name(), suffix) {
+				excluding = true
+				break
+			}
+		}
```
ilyam8: this is getting complicated, i suggest to simplify, it looks like it is time to wrap []string (includeSuffix, excludeSuffix) to its own type and add methods to it, after it it will be somewthing like that

```go
for _, fi := range fis {
    if !fi.Mode().IsRegular() || isBadFileName(fi.Name()){
        continue
    }

    if !dir.inlcudeSuffic.match(fi.Name()) || dir.excludeSuffic.match(fi.Name()) {
        continue
    }
    ...
    ...  
```

nested if/else with break is hard to read/follow, please simplify

ilyam8: ^^ it is likely i am wrong with logic, please check it, code snippet above is just to show the idea

ilyam8: i think it should be

```go
if dir.matchConfig(fi.Name()) { ... }
```

because it should match every file if both includeSuffinx and excludeSuffix is nil

kovetskiy: I think two dead simple loops have much less cognitive load than having two separated types which will be used in one place. Moreover, adding separated types will make logic fragmented.

modules/dnsmasq_dhcp/autodetection.go
kovetskiy: I've added `isValidFileName` and pushed commit with force

modules/dnsmasq_dhcp/autodetection.go
ilyam8:
```go
for _, fi := range fis {
    if !fi.Mode().IsRegular() {
        continue
    }

    isBadFileName(fi.Name()) {
         continue
     }

     if !dir.matchConfig(fi.Name())  {
         continue
     }
```

vs

```go
	for _, fi := range fis {
		if !fi.Mode().IsRegular() {
			continue
		}

		// We copy the dnsmasq's logic
		//
		// /* ignore emacs backups and dotfiles */
		// if (len == 0 ||
		//     ent->d_name[len - 1] == '~' ||
		//     (ent->d_name[0] == '#' && ent->d_name[len - 1] == '#') ||
		//     ent->d_name[0] == '.')
		//    continue;

		name := fi.Name()
		if strings.HasSuffix(name, "~") ||
			(strings.HasPrefix(name, "#") && strings.HasSuffix(name, "#")) ||
			strings.HasPrefix(name, ".") {
			continue
		}

		if len(dir.includeSuffix) > 0 {
			including := false
			for _, suffix := range dir.includeSuffix {
				if strings.HasSuffix(fi.Name(), suffix) {
					including = true
					break
				}
			}

			if !including {
				continue
			}
		}

		excluding := false
		for _, suffix := range dir.excludeSuffix {
			if strings.HasSuffix(fi.Name(), suffix) {
				excluding = true
				break
			}
		}

		if excluding {
			continue
		}

		configs = append(configs, filepath.Join(dir.path, fi.Name()))
	}

```

ilyam8: in 1 case we have only logic, in 2nd logic + details. You can add `matchConfig` to `configDir` struct. Whatever you want.

kovetskiy: https://github.com/netdata/go.d.plugin/pull/239#discussion_r301149374

ilyam8: yes, i saw, nice, thanks :+1:

ilyam8: Oh, please dont force push on every commit, we will squash them during merge

kovetskiy: I pushed since you insist, but I don't like that changeset because it makes harder to ensure that the module follows dnsmasq's logic correctly.

ilyam8: Why is it harder now? What is changed? We moved implementation details to its own method which can be tested afther those changes.

ilyam8: So your point that before it was copy of `c` code and now it is not?

kovetskiy: I found the code from the first iteration much more convenient for side-by-side checking logic with dnsmasq's code.

kovetskiy: Let me know if you have any questions/issues with the code.

ilyam8: Ok, i got your point.

modules/dnsmasq_dhcp/autodetection.go
```diff
@@ -46,21 +60,96 @@ func (d *DnsmasqDHCP) findDHCPRanges() ([]string, error) {
...
+		dirFiles, err := dir.findConfigs()
+		if err != nil {
+			d.Warningf("error during configuration dir %q scanning: %v", dir.path, err)
+		}
+
+		includeFiles = append(includeFiles, dirFiles...)
+	}
+
+	for _, file := range includeFiles {
+		files := d.findConfigs(file)
```
ilyam8: i think we need to avoid *infinite* recursion here (it is not infinite, i know)

that is what i get, fatal error doesnt look good
```
[ DEBUG ] dnsmasq_dhcp[dnsmasq_dhcp] autodetection.go:126 664436 opening/etc/dnsmasq.conf
[ DEBUG ] dnsmasq_dhcp[dnsmasq_dhcp] autodetection.go:126 664437 opening/etc/dnsmasq.conf
[ DEBUG ] dnsmasq_dhcp[dnsmasq_dhcp] autodetection.go:126 664438 opening/etc/dnsmasq.conf
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

modules/dnsmasq_dhcp/autodetection.go
```diff
@@ -46,21 +60,96 @@ func (d *DnsmasqDHCP) findDHCPRanges() ([]string, error) {
...
 }
 
-func findConfigurations(confDir string) ([]string, error) {
-	fis, err := ioutil.ReadDir(confDir)
+// findConfigs recursively finds and reads configuration files respecting
+// conf-file and conf-dir directives.
+// findConfigs is tolerant to IO errors and finds as maximum config
+// files as possible, therefore if an error occurrs during scanning process,
+// it will be just logged with warning severity
+func (d *DnsmasqDHCP) findConfigs(confPath string) []string {
```
ilyam8: `findConfigs` returns configs, right? but it doesnt include entry config (`confPath string`). Looks a little bit lame to me, not sure. What do you think?

having it this way we need to

```
	configs := d.findConfigs(d.ConfPath)
	if err != nil {
		d.Warningf("error during configuration dir scanning : %v", err)
    }
	configs = append([]string{d.ConfPath}, configs...)
```

^^ looks like a workaround or smth

kovetskiy: Didn't touch that logic though

https://github.com/netdata/go.d.plugin/pull/239/files#diff-daece282ff93570d72a99d7ffa78dfd5L20

ilyam8: i know, i know.

it was

```
findConfigurations(confDir string) ([]string, error)
```

which is ok (but not good ofc), if finds configs in dir.

we can fix it later.

One things i noticed, your implementation ignores `ConfDir`.

ilyam8: Well, maybe we need to get rid of `ConfigDir`

All we need is the entry config

```
CONFIG FILE
       At startup, dnsmasq reads /etc/dnsmasq.conf, if it exists. (On FreeBSD, the file is /usr/local/etc/dnsmasq.conf ) (but see the -C and -7 options.) The format of this file consists of one option per line, exactly as the long options detailed in the OPTIONS section but without the
       leading "--". Lines starting with # are comments and ignored. For options which may only be specified once, the configuration file overrides the command line.  Quoting is allowed in a config file: between " quotes the special meanings of ,:. and # are removed and  the  following
       escapes are allowed: \\ \" \t \e \b \r and \n. The later corresponding to tab, escape, backspace, return and newline.
```

ilyam8: On debian it reads `/etc/default/dnsmasq`


from
> /etc/init.d/dnsmasq

```
NAME=dnsmasq
DESC="DNS forwarder and DHCP server"

# Most configuration options in /etc/default/dnsmasq are deprecated
# but still honoured.
ENABLED=1
if [ -r /etc/default/$NAME ]; then
        . /etc/default/$NAME
fi

```

ilyam8: ok, it seems `/etc/default/dnsmasq` doesnt work, i found in README in `dnsmasq.d`

```
# All files in this directory will be read by dnsmasq as 
# configuration files, except if their names end in 
# ".dpkg-dist",".dpkg-old" or ".dpkg-new"
#
# This can be changed by editing /etc/default/dnsmasq
```

but, as i said it doesnt work, so we can ignore it.

ilyam8: i finally read `man dnsmasq` lmao

```
       -C, --conf-file=<file>
              Specify a different configuration file. The conf-file option is also allowed in configuration files, to include multiple configuration files. A filename of "-" causes dnsmasq to read configuration from stdin.

       -7, --conf-dir=<directory>[,<file-extension>......],
              Read all the files in the given directory as configuration files. If extension(s) are given, any files which end in those extensions are skipped. Any files whose names end in ~ or start with . or start and end with # are always skipped. If the extension starts with * then
              only files which have that extension are loaded. So --conf-dir=/path/to/dir,*.conf loads all files with the suffix .conf in /path/to/dir. This flag may be given on the command line or in a configuration file. If giving it on the command line, be sure to escape  *  charac‐
              ters.

```

it looks like we need both `config_path` and `config_dir` since both of them can be specified via command line.

PR #242, opened 2019-07-17
scaleio: Remove numOfScsiInitiators

mihkelvain:
It is not in use and breaks ScaleIO/VxFlex v3 monitoring

Fixes: netdata/netdata#6475

### DIFF
--- a/modules/scaleio/queries.go
+++ b/modules/scaleio/queries.go
@@ -62,7 +62,6 @@ var (
         "numOfMappedToAllVolumes",
         "numOfProtectionDomains",
         "numOfRfcacheDevices",
-        "numOfScsiInitiators",
         "numOfSdc",
         "numOfSds",
         "numOfSnapshots",
--- a/modules/scaleio/queries.go
+++ b/modules/scaleio/queries.go
@@ -62,6 +62,9 @@ var (
         "numOfMappedToAllVolumes",
         "numOfProtectionDomains",
         "numOfRfcacheDevices",
+	// Starting from version 3 of ScaleIO/VxFlex API numOfScsiInitiators property is removed.
+	// Reference: VxFlex OS v3.x REST API Reference Guide.pdf
+	// "numOfScsiInitiators",
         "numOfSdc",
         "numOfSds",
         "numOfSnapshots",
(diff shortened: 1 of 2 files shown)


modules/scaleio/queries.go
```diff
@@ -62,7 +62,6 @@ var (
         "numOfMappedToAllVolumes",
         "numOfProtectionDomains",
         "numOfRfcacheDevices",
-        "numOfScsiInitiators",
```
ilyam8: i think better to comment out it and add a comment with a reason

ilyam8:
And one more minor thing, please update [scaleio readme](https://github.com/netdata/go.d.plugin/tree/master/modules/scaleio#scaleio) `tested on` section

```md
Module was tested on:
 - OS v2.6.1.1_113, REST API v2.5
 - your version
```

modules/scaleio/queries.go
mihkelvain: Good point. Modified

ilyam8:
Do you know what is your rest api version? 

i see there is a method to get version 

https://github.com/netdata/go.d.plugin/blob/c9b76c2e1e5bd5890304f83af8888ce5feb5ef8c/modules/scaleio/client/client.go#L69-L79

but for some reason i didnt use it in check to log the version

ilyam8:
you can check rest api version with a simple get request, you dont need to be authenticated to make it

url path is

https://github.com/netdata/go.d.plugin/blob/c9b76c2e1e5bd5890304f83af8888ce5feb5ef8c/modules/scaleio/client/endpoints.go#L17

mihkelvain:
GET /api/version returns "3.0"

ilyam8:
Ok, add it to the readme to. Like in [this my comment](https://github.com/netdata/go.d.plugin/pull/242#issuecomment-512134015). Not sure this is needed, but if it tested on this version i want to have it.

ilyam8:
@mihkelvain thanks :+1:

ilyam8:
Comment out was a bad advice from me, that query should be in json format, i fixed the problem in https://github.com/netdata/go.d.plugin/commit/627a306d934fec617bfdfc1f5cbd869c68ae8931

PR #252, opened 2019-08-16
netdata/go.d: Attempt to fix build-id errors during RPM build

paulkatsoulakis:
Attempt to fix build-id errors during RPM build as per https://github.com/rpm-software-management/rpm/issues/367 recommendation


paulkatsoulakis:
cancelling - @ilyam8 identified some concerns around this solution, we 'll follow the other workaround that goes on the RPM spec file side

PR #253, opened 2019-08-21
php-fpm

nrwiersma:
Hi,

I have ported php-fpm from your python module.

Please let me know if there is anything i missed in the process.

### DIFF
--- a/modules/phpfpm/client.go
+++ b/modules/phpfpm/client.go
@@ -1,0 +1,80 @@
+package phpfpm
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/netdata/go.d.plugin/pkg/web"
+)
+
+type status struct {
+	Active    int64  `json:"active processes" stm:"active"`
+	MaxActive int64  `json:"max active processes" stm:"maxActive"`
+	Idle      int64  `json:"idle processes" stm:"idle"`
+	Requests  int64  `json:"accepted conn" stm:"requests"`
+	Reached   int64  `json:"max children reached" stm:"reached"`
+	Slow      int64  `json:"slow requests" stm:"slow"`
+	Processes []proc `json:"processes"`
+}
+
+type proc struct {
+	PID      int64   `json:"pid"`
+	State    string  `json:"state"`
+	Duration int64   `json:"request duration"`
+	CPU      float64 `json:"last request cpu"`
+	Memory   int64   `json:"last request memory"`
+}
+
+type client struct {
+	client *http.Client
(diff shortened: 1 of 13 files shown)


ilyam8:
@nrwiersma thx! will review the PR tomorrow 😉

modules/phpfpm/client.go
```diff
@@ -0,0 +1,78 @@
...
+		resp.Body.Close()
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("%s returned HTTP status %d", req.URL, resp.StatusCode)
+	}
+
+	s := &status{}
+	if err := c.dec(resp.Body, s); err != nil {
+		return nil, fmt.Errorf("error parsing response")
```
ilyam8: > fmt.Errorf("error parsing response")

forgot to add the error

modules/phpfpm/phpfpm.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	data := stm.ToMap(status)
+	if len(status.Processes) == 0 {
+		return data
+	}
+
+	statProcesses(data, status.Processes, "ReqDur", func(p proc) int64 { return p.Duration })
+	statProcesses(data, status.Processes, "ReqCpu", func(p proc) int64 { return int64(p.CPU) })
+	statProcesses(data, status.Processes, "ReqMem", func(p proc) int64 { return p.Memory })
```
ilyam8: please move collect logic to the `collect.go`

https://github.com/netdata/go.d.plugin/blob/e2059a14dc50d2c4acd86e64dfde4b5f1f99a15c/modules/vsphere/vsphere.go#L187-L199

modules/phpfpm/decode.go
```diff
@@ -0,0 +1,125 @@
...
+	"strings"
+)
+
+type decoder func(r io.Reader, s *status) error
+
+func decodeJSON(r io.Reader, s *status) error {
+	return json.NewDecoder(r).Decode(s)
+}
+
+func decodeText(r io.Reader, s *status) error {
```
ilyam8: overall decodeText func is hard to follow, how about to split it into several smaller ones?

modules/phpfpm/phpfpm.go
nrwiersma: Is there any particular reason for this? From a Go perspective, this is not needed or probably wanted. You are merely splitting code a) into a function and b) into a file when the logic is in fact really simple.

ilyam8: The reason is having `collect.go` as an entry point for collecting logic in all modules

PR #257, opened 2019-08-27
[WIP] Redis collector

fracasula:
Hey guys, this is to migrate the Redis collector from Python here to the `go.d.plugin` codebase.

It's almost done even though I think I'll spend some more time doing manual a few more manual tests. In the meantime I got a few questions for you.

1. It seems to me that the `bgsave_now` chart on python isn't a "line" chart, the last element in that list is "absolute" but there no such chart type (i.e. line|area|stacked) so for now I've put "line" in there, let me know if it doesn't make sense
   * https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/redis/redis.chart.py#L118
   * https://github.com/fracasula/go.d.plugin/blob/redis-collector/modules/redis/charts.go#L162
2. Should I change this back to `Disabled: true`?
   * https://github.com/fracasula/go.d.plugin/blob/redis-collector/modules/redis/redis.go#L15
3. I still got one chart left to convert which is the `keys_pika` one but I'm not sure yet about what it does. My guess is that I should check whether this [Pika](https://github.com/Qihoo360/pika/) is running instead of pure Redis and act accordingly?


fracasula:
Note: I still have to test it with a password protected Redis instance.

modules/redis/charts.go
```diff
@@ -0,0 +1,190 @@
...
+		Units: "KiB",
+		Fam:   "memory",
+		Ctx:   "redis.memory",
+		Type:  module.Line,
+		Dims: Dims{
+			{
+				ID:   "used_memory",
+				Name: "total",
+				Div:  1024,
+			},
```
ilyam8: lets have one line dimension, i would prefer to have them consistent across all charts

modules/redis/charts.go
```diff
@@ -0,0 +1,190 @@
...
+	},
+	{
+		ID:    "connections",
+		Title: "Connections",
+		Units: "connections/s",
+		Fam:   "connections",
+		Ctx:   "redis.connections",
+		Type:  module.Line,
+		Dims: Dims{
+			{ID: "total_connections_received", Name: "received", Algo: module.Incremental, Mul: 1},
```
ilyam8: `Mul: 1` is default, no reason to explicitly use it

modules/redis/charts.go
```diff
@@ -0,0 +1,190 @@
...
+	},
+	{
+		ID:    "clients",
+		Title: "Clients",
+		Units: "clients",
+		Fam:   "connections",
+		Ctx:   "redis.clients",
+		Type:  module.Line,
+		Dims: Dims{
+			{ID: "connected_clients", Name: "connected", Mul: 1},
```
ilyam8: same, 1 is default

modules/redis/charts.go
```diff
@@ -0,0 +1,190 @@
...
+			{ID: "total_net_output_bytes", Name: "out", Algo: module.Incremental, Mul: -8, Div: 1000},
+		},
+	},
+	{
+		ID:    keysRedisChartId,
+		Title: "Keys per Database",
+		Units: "keys",
+		Fam:   "keys",
+		Ctx:   "redis.keys",
+		Type:  module.Line,
```
ilyam8: `module.Line` is the default, please remove (true for all charts with type line charts)

modules/redis/redis.go
```diff
@@ -0,0 +1,267 @@
...
+
+	err = parseMetrics(res.(string), r.charts, metrics, &r.bgSaveTime, r.UpdateEvery)
+	if err != nil {
+		r.Errorf("Got error while parsing metrics: %v", err)
+		r.Debugf("INFO: %+v", res)
+	}
+
+	r.Debugf("Metrics: %+v", metrics)
+
+	return metrics
```
ilyam8: Please move collect logic to `collect.go`. Reason - it is nice to have `collect.go` as an entry point for collecting logic for all modules.


https://github.com/netdata/go.d.plugin/blob/02b5f76a43e0f8cb69e936f3a33560797ae2b178/modules/zookeeper/zookeeper.go#L95-L105

modules/redis/redis_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/netdata/go-orchestrator/module"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
```
ilyam8: Please add test for `Init` and `Check`. I see only tests for `happy path`, would be nice to have some fail scenarios

modules/redis/redis.go
```diff
@@ -0,0 +1,267 @@
...
+	"total_net_input_bytes", "total_net_output_bytes",
+	"evicted_keys",
+	"total_connections_received", "rejected_connections",
+	"connected_clients", "blocked_clients",
+	"connected_slaves",
+	"rdb_changes_since_last_save",
+	"uptime_in_seconds",
+}
+
+func parseMetrics(
```
ilyam8: i think better to have the function as method, it doesnt look good to me how it is implemented right now:
 - a lot of input parameters
 - output of function actually is input (changes metrics and charts)
 - function name is parse, but it does more that it states - it parses, updates metric map, updates charts, etc.

What do you think?

ilyam8:
> My guess is that I should check whether this Pika is running instead of pure Redis and act accordingly?

Yes

modules/redis/charts.go
fracasula: Sure, no problem with sticking with the defaults here and in the other places (i.e. `Mul`, `Line`...) :+1:

modules/redis/redis.go
fracasula: Sounds good to me :+1:

modules/redis/redis_test.go
fracasula: Yeah I'm definitely not done with the tests here but I'd like to spend more time going through the framework you guys wrote before trying to write a test for `Init` or `Check`. I don't want to end up just testing implementation details like [here](https://github.com/netdata/go.d.plugin/blob/master/modules/rabbitmq/rabbitmq_test.go#L20) or [here](https://github.com/netdata/go.d.plugin/blob/master/modules/rabbitmq/rabbitmq_test.go#L24) for example. I'll try to see if I can come up with some tests for those methods that bring more value. 

Also since we're talking about tests it's starting to be a bit of a burden having to test this against a real Redis/Pika combination. An integration test could bring lots of value here I think. I noticed you guys have a [docker-compose](https://github.com/netdata/go.d.plugin/blob/master/docker-compose.yml), could we potentially use that one to bring a few more containers up? Thoughts?

modules/redis/charts.go
fracasula: Sure, I can do that. Gotta be honest though, first time since I started working with Go that I see a comment in a code review about formatting. It's fine it's just that I usually prefer to see more constructive comments and I thought I'd point that out.

I hope you don't mind the criticism :slightly_smiling_face:

modules/redis/redis.go
fracasula: I agree on almost everything, let's break this down.

* output of function actually is input
   * I sometimes do that to avoid gc pressure but since the metrics map would escape to the heap anyway in the `Collect()` method we wouldn't avoid it anyway so I can definitely make `parseMetrics` initialize and return the map instead of having it as an input
* a lot of input parameters / function name is parse, but it does more that it states
   * I think both of these points you make could get down to the same thing, `parseMetrics` has too many responsibilities. I'd like to buy some time here and see how it looks like once it lives in the `collect.go` file and I add the Pika support. With Pika it should grow enough in complexity to give me an idea of how to better split this up.

About it being a method instead of a function I think I prefer it to be a function as it simplifies the tests. By having an unexported function wrapped in an exported method we could still expose a really nice API to whoever uses our package and at the same time we have a small unit (the private unexported function) that is super easy to test. More on this approach [here](https://stackoverflow.com/a/48206430/828366) if you can spare the time I'd like to know what you think about it.

modules/redis/charts.go
cosmix: I'm ok with stating things explicitly. Explicit > Implicit :)

fracasula: That would be my personal preference as well, I think it makes things more readable but in this case I don't mind either way. Up to you guys.

ilyam8: I agree that explicit > implicit in general, but not in this case.

>  I think it makes things more readable

In my opinion we will get more noise then readability, so it is not worth it.

modules/redis/redis_test.go
ilyam8: Yes, those tests you mentioned are not good. 

> An integration test could bring lots of value

Impossible to argue 😄

modules/redis/redis.go
ilyam8: I think we need to use interface as `client` instead of `*redis.Client` and mock it during tests, otherwise you have to test `parseMetrics`. Testing parseMetrics != testing Collect (we are testing implementation details).

ilyam8: > a really nice API to whoever uses our package

it is not a library, our modules are specific to netdata, their output has no sense outside netdata (unfortunately)

modules/redis/charts.go
ilyam8: no, all is good

modules/redis/redis.go
ilyam8: oh, i see the same problem in the your approach 

> To summarize: test the unexported functions instead of testing the exported ones!

Maybe i am missing something, for me the problem is testing the details instead of logic, coupling logic with implementation.

PR #264, opened 2019-09-05
fix context on springboot2

Wing924:


ilyam8:
it looks like https://github.com/netdata/go.d.plugin/pull/263

ilyam8:
dup of #263

PR #270, opened 2019-09-17
Add Current State Duration to HTTPcheck

ReillyProcentive:
This adds the current state duration chart (from portcheck) to httpcheck as well. We needed this because we wanted to be able to alert on flapping HTTP checks (IE the HTTP checks are allowed to fail every so often but if they fail repeatedly in a time frame then there might be an underlying issue).


ilyam8:
thanks @ReillyProcentive

PR #275, opened 2019-10-01
rabbitmq collect vhosts

ilyam8:
rel: netdata/netdata#6686


mahamafzar:
`/node` endpoint is invalid, `/nodes` is valid

ilyam8:
thx @mahamafzar i fixed it in #279

PR #276, opened 2019-10-02
rabbitmq: family per vhost

ilyam8:
ssia


modules/rabbitmq/collect.go
```diff
@@ -116,10 +117,18 @@ func (r *RabbitMQ) parse(mx *metrics) map[string]int64 {
...
 	}
 
 	return ms
 }
+
+func cleanVhostName(name string) string {
+	if name == "/" {
+		return name
+	}
+	return strings.TrimPrefix(name, "/")
```
Wing924: vhost may be /foo/bar

Wing924: we should allow / in vhost name but don’t use it in chart id

ilyam8: So whats your suggestion?

I think it would be better to have `vhost /foo/bar` family, then `foo_bar`? Do we really need to change vhost name?

ilyam8: > we should allow / in vhost name but don’t use it in chart id

what about dimID?

PR #277, opened 2019-10-03
Fix markdown syntax HDFS/Zookeeper module pages

joelhans:
Hey @ilyam8, I noticed that the READMEs for the new HDFS and Zookeeper modules, once they get generated by our docs site, don't render correctly. The unordered list gets turned into a big, messy paragraph.

Here's a picture of how the HDFS page looks right now:

image

I cleaned up the syntax and tested it out on a local build of the docs site and things work as expected after these changes.

Not sure _why_ GitHub and mkdocs render Markdown quite so differently, but that's for another time...

### DIFF
--- a/modules/hdfs/README.md
+++ b/modules/hdfs/README.md
@@ -1,51 +1,54 @@
 # hdfs
 
-This module will monitor one or more [`Hadoop Distributed File System`](https://hadoop.apache.org/docs/r1.2.1/hdfs_design.html) (HDFS) nodes depending on configuration.
+This module monitors one or more [`Hadoop Distributed File
+System`](https://hadoop.apache.org/docs/r1.2.1/hdfs_design.html) (HDFS) nodes, depending on your configuration.
 
-It access HDFS metrics over `Java Management Extensions` (JMX) through the web interface of an HDFS daemon.
+Netdata accesses HDFS metrics over `Java Management Extensions` (JMX) through the web interface of an HDFS daemon.
 
 **Requirements:**
- * `hdfs` node with accessible `/jmx` endpoint
+
+-   `hdfs` node with accessible `/jmx` endpoint
 
 It produces the following charts for `namenode`:
-  - Heap Memory in `MiB`
-  - GC Events in `events/s`
-  - GC Time in `ms`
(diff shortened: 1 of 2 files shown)


ilyam8:
> Not sure why GitHub and mkdocs render Markdown quite so differently

Well, i use JB(jet brains, ide) markdown plugin.

Screenshot 2019-10-04 at 00 15 47

___

i think all READMEs are broken then

ilyam8:
thx @joelhans 👍

ilyam8:
@joelhans i found online [markdown linter](https://dlaa.me/markdownlint/)

i posted there fixed version of `hdfs` readme and it shows a lot of errors.

Almost all of them

> Spaces after list markers [Expected: 1; Actual: 3]

I kind of agree with it

```markdown
It produces the following charts for `namenode`:

-   Heap Memory in `MiB`
-   GC Events in `events/s`
```

Why we use so many spaces after list marker?

joelhans:
The Markdown linter we have in the core repo (https://github.com/netdata/netdata/blob/master/.remarkrc.js) is what demands those spaces, from the `remark-lint-list-item-indent` package.

Remark was set like that from before I began, so I've just kept up enforcing it as our standard.

We should probably set up a linter on this repo, too—I always forget about this one. Why is this separate from the core repo, anyway?

ilyam8:
> Why is this separate from the core repo

it doesnt make much sense to have it in netdata repo, it would only make development harder, this is an external thing to netdata. This tool does not require netdata to be running/testing.
Netdata has no golang as a dependency to compile it, installer just downloads appropriate version from the [releses page](This tool does not require netdata to be running/testing.)

ilyam8:
> Remark was set like that from before I began, so I've just kept up enforcing it as our standard.

ok, it is a matter of opinion but 3 spaces looks not good to me. At least it doesnt make text clear for me.

joelhans:
I agree it's not great. Markdown in general is kind of a pain. But considering we just went through a ton of .md files to give lists that indent, it'll be tough to convince me to go back :stuck_out_tongue: 

And thanks for the explanation about why this is separate. That makes a lot of sense, actually.

PR #283, opened 2019-10-25
Convert check.ServiceTags []string to string and added a new dimension

ebarriosjr:


ebarriosjr:
Sorry, didn't mean to push this upstream yet. Need to test it before doing this.

ilyam8:
Np @ebarriosjr

PR #318, opened 2020-01-09
Feat(nvidia-smi): rewrite nvidia-smi module with native C API

Ehekatl:
not sure if it's acceptable...
this PR enables CGO, and embedded 3rd party lib directly under pkg.
I couldn't use go mod from source because I made some changes there (to fix nil point issue, etc 
Also the chart names are slightly different from python version.

Other than that everything works well, I've being use this for quite long.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -5,7 +5,7 @@ WORKDIR /workspace
 
 ENV GOOS=linux
 ENV GOARCH=amd64
-ENV CGO_ENABLED=0
+ENV CGO_ENABLED=1
 
 ADD go.mod go.sum ./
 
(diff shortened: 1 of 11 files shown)


tycho:
Nice! This looks way better than my slow/kludgy Python garbage.

ilyam8:
hi @Ehekatl 

> this PR enables CGO

that is what we are trying to avoid, i removed nvidia and oracledb modules because of that

https://github.com/netdata/go.d.plugin/pull/204


> I made some changes there

Why do not contribute to the upstream?

Ehekatl:
Ya I know @ilyam8 
That's why I didn't expect to merge this. Just add it here in case someone need it, feel free to close

ilyam8:
@Ehekatl you could write it as a standalone plugin

> https://github.com/netdata/go-orchestrator/blob/master/examples/simple/main.go

and add a link here

> https://docs.netdata.cloud/collectors/collectors/#third-party-plugins

PR #324, opened 2020-01-22
Add note about required mysql permissions

### LINKED ISSUE #323: MySQL module permissions
Docs should specify required permissions for the user which is specified in configuration, maybe even the example, like python plugin is https://docs.netdata.cloud/collectors/python.d.plugin/mysql/

I assume `USAGE, REPLICATION CLIENT` should be enough?

arr-dev:
Closes #323

### DIFF
--- a/modules/mysql/README.md
+++ b/modules/mysql/README.md
@@ -4,6 +4,18 @@
 
 This module will monitor one or more `MySQL` servers depending on configuration.
 
+## Requirements
+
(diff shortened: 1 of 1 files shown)


ilyam8:
Thx @arr-dev 

How about to move it to the `Requirements` section, like

https://github.com/netdata/go.d.plugin/tree/master/modules/unbound#requirements

PR #339, opened 2020-02-22
go.mod: add support for freebsd/arm64

dmgk:
Go 1.14 will add support for freebsd/arm64, `golang.org/x/sys` and `golang.org/x/net` that are pulled as implicit dependencies are too old to support it.

Add explicit indirect deps to `golang.org/x/net@83d349e8ac1a` and `golang.org/x/sys@33540a1f6037` which are minimal versions that support freebsd/arm64

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,8 @@ require (
 	github.com/prometheus/prometheus v2.5.0+incompatible
 	github.com/stretchr/testify v1.4.0
 	github.com/vmware/govmomi v0.22.1
+	golang.org/x/net v0.0.0-20191027093000-83d349e8ac1a // indirect
+	golang.org/x/sys v0.0.0-20191026070338-33540a1f6037 // indirect
 	gopkg.in/yaml.v2 v2.2.7
 	layeh.com/radius v0.0.0-20190322222518-890bc1058917
 )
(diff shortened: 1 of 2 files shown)


ilyam8:
Hi @dmgk 👋 First of all thank you for the contibution!


I am ok to merge it, but could you exaplain why we need it today?

If i understand it right `support for freebsd/arm64` will be automatically added when 1.14 officialy released (and `go mod tidy`)

dmgk:
@ilyam8 although go1.14 itself will support freebsd/arm64, third-party deps that `netdata/go.d.plugin` references do not, hence the need to update their versions

ilyam8:
@dmgk 

Probably i am missing something, see

> https://github.com/golang/go/wiki/Modules#how-to-prepare-for-a-release

> Run `go mod tidy` to possibly prune any extraneous requirements (as described here) and also ensure your current go.mod reflects all possible build tags/OS/architecture combinations


So, when 1.14 is released, i do `go mod tidy` and that is it. Am i correct?

ilyam8:
ok, i was wrong here https://github.com/netdata/go.d.plugin/pull/339#issuecomment-590062144

> pulled as implicit dependencies are too old to support it

this 🙈 

Thx @dmgk i added `freebsd/arm64` in #344

it works

```cmd
366 ilyam:go.d.plugin (bump_go1.14)$ ./hack/go-build.sh freebsd/arm64
Building binaries for version: v0.15.0-18-g8f34c5d
Building for os/arch freebsd/arm64
```


Thx again

PR #348, opened 2020-02-27
Added ReviewDog Go linting checks.

Ferroin:
We mistakenly added these to the agent repo instead of this one, add them here instead.

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -1,0 +1,14 @@
+name: reviewdog
+on: [pull_request]
+jobs:
+  golangci-lint:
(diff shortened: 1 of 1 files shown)


ilyam8:
i think i have all those checks in circleci

ilyam8:
anyway

ilyam8:
was wrong
https://github.com/netdata/go.d.plugin/pull/350/checks?check_run_id=472485497

good good

Ferroin:
As mentioned in the standup, the plan is to move to GHA as much as possible for CI. Depending on how things end up, we'll probably even ben able to do the FreeBSD builds using it.

PR #376, opened 2020-05-07
Add frontmatter to fix titles in docs site

joelhans:
Connected to netdata/netdata-learn-docusaurus#108 and netdata/netdata#8936. This should be merged first, then 8936, and then 108.

This PR adds frontmatter to all the README files so that they show the correct title in the navigation and on individual pages in the docs. I've also changed their names in the sidebar to reflect the service/app being monitored, not the module's name.


ilyam8:
should we update [plugin readme](https://github.com/netdata/go.d.plugin#godplugin)?

PR #401, opened 2020-07-15
prometheus collector

ilyam8:
fixes: netdata/netdata#9470

This PR adds prometheus module. It is basically visualization of prometheus metrics.

**The idea is questionable**, but will see. 

It requires only `url` parameter, it doesnt have any other options, because it is not clear what we need.

Some things are hardcoded and it is deliberately:
- max chart per metric
- desired number of dimensions per chart
- max number of dimensions per chart

Charts:

- 2 charts per every summary and histogram. Chart id: all labels except `le`/`quantile`, dimension id: all labels. Charts: `incremental` + `percentage-of-incremental-row`.
- 1 or more charts per every gauge, counter. Dimension id: all labels. If number of dimensions > desired number of dimensions per chart we create several charts.


config/go.d/prometheus.conf
```diff
@@ -0,0 +1,160 @@
+# netdata go.d.plugin configuration for prometheus
+#
+# This file is in YaML format. Generally the format is:
```
prologic, line 3:
```suggestion
# This file is in YAML format. Generally the format is:
```

config/go.d/prometheus.conf
```diff
@@ -0,0 +1,160 @@
...
+#  - tls_key
+#    Client tls key.
+#    Syntax:
+#      tls_key: path/to/key.pem
+#
+#
+# [ JOB defaults ]:
+#  timeout: 5
+#  method: GET
+#  not_follow_redirects: no
```
prologic, line 139: Don't like this key. Suggestion: `follow_redirects: yes`

config/go.d/prometheus.conf
```diff
@@ -0,0 +1,160 @@
...
+# param1: value1
+# param2: value2
+#
+# Currently supported global parameters:
+#  - update_every
+#    Data collection frequency in seconds. Default: 1.
+#
+#  - autodetection_retry
+#    Re-check interval in seconds. Attempts to start the job are made once every interval.
+#    Zero means not to schedule re-check. Default: 0.
```
prologic, line 25: I don't understand this at all. Can you explain how this works? Then maybe we can reword this and use a better name?

config/go.d/prometheus.conf
```diff
@@ -0,0 +1,160 @@
...
+# [ GLOBAL ]
+# These variables set the defaults for all JOBs, however each JOB may define its own, overriding the defaults.
+#
+# The GLOBAL section format:
+# param1: value1
+# param2: value2
+#
+# Currently supported global parameters:
+#  - update_every
+#    Data collection frequency in seconds. Default: 1.
```
prologic, line 21: Let's not make the default `1s` because hitting a `/metrics` endpoint that hard will be a) unnecessary HTTP traffic and load and b) noisey in server logs. Good default: `5s`

config/go.d/prometheus.conf
```diff
@@ -0,0 +1,160 @@
...
+# priority: 70000
+#
+#
+# [ JOBS ]
+#jobs:
+#  - name: node_exporter
+#    url: http://127.0.0.1:9100/metrics
+#
+#  - name: win10
+#    url: http://203.0.113.0:9182/metrics
```
prologic, line 160: I like the simplicity of creating job configurations.

What you need here as well is some way to relabel metrics as you scrape them and apply filters/replacements. Can be done in a future set of PR(s) I _think_ the config/spec is open to expansion?

config/go.d/prometheus.conf
ilyam8: well, i need something that defaults to _follow_ if not set 😄 

I just embed web struct

https://github.com/netdata/go.d.plugin/blob/359f7334899f691a26701d57b0075471b8b98a80/modules/wmi/wmi.go#L44-L53

And it has all web related options with ok defaults

https://github.com/netdata/go.d.plugin/blob/359f7334899f691a26701d57b0075471b8b98a80/pkg/web/request.go#L12-L22

https://github.com/netdata/go.d.plugin/blob/359f7334899f691a26701d57b0075471b8b98a80/pkg/web/client.go#L15-L20

config/go.d/prometheus.conf
ilyam8: This came from python plugin, same option there. 

How it works:

 - create job
 - **before run it**: try to check (module interface has `Check() bool`)
 - if check is success - run
 - if check is fail - drop the job

Sometimes you dont want to drop the job after the check and want the plugin to keep checking it. This options allows you to configure re-check interval.

This is related code from the plugin

https://github.com/ilyam8/go-orchestrator/blob/e21d211f5a17d82ee8331989cf96cee5cafd4d8e/job/build/build.go#L238-L263

If we need to retry (retry check) the job we create retry task

```go
go retryTask(ctx, m.retryCh, cfg)
```

which is wait for `autodetection_retry` and place it in the job build queue again (instead of dropping the job)

```go
func retryTask(ctx context.Context, in chan<- confgroup.Config, cfg confgroup.Config) {
	timeout := time.Second * time.Duration(cfg.AutoDetectionRetry())
	t := time.NewTimer(timeout)
	defer t.Stop()

	select {
	case <-ctx.Done():
	case <-t.C:
		select {
		case <-ctx.Done():
		case in <- cfg:
		}
	}
}
```

The whole process is https://github.com/ilyam8/go-orchestrator/blob/master/job/build/build.go

---


> Then maybe we can reword this

It was suggested by Costa, see https://github.com/netdata/netdata/issues/2757

config/go.d/prometheus.conf
ilyam8: Agree, in fact i made it `5s` by default

https://github.com/ilyam8/go.d.plugin/blob/46e45d32bc84f4193f5a40f73f4a723276b4cbd6/modules/prometheus/prometheus.go#L12-L20

Forgot to change it in the config

config/go.d/prometheus.conf
ilyam8: Well, it could have sense if netdata supported labels - dont forget we have only `charts` and `dimenstions`. 

`metric_name{labelA="valueA",labelB="valueB"}` => (chart dimension) `metric_name|labelA=valueA,labelB=valueB`

> I think the config/spec is open to expansion?

Yes, since it has 0 options (dont count connection related) it is open to expansion 😄

config/go.d/prometheus.conf
prologic: Cool thanks ;)

config/go.d/prometheus.conf
prologic: THanks!

config/go.d/prometheus.conf
prologic: Thanks!

prologic:
Thanks for the answers to my dumb questions @ilyam8  :)

PR #404, opened 2020-07-21
module/mysql: feature parity with py version

ilyam8:
- [X] Fixes: netdata/netdata#9448
- [X] Fixes: netdata/netdata#9006
- [X] Fixes: netdata/netdata#9041
- [x] Fixes: netdata/netdata#8785

Most of the added SLOC is test data (7k).


modules/mysql/mycnf.go
```diff
@@ -0,0 +1,77 @@
+package mysql
```
prologic, line 1: Why not use [one of these libraries](https://github.com/search?q=language:Go%20my%20cnf&ref=opensearch)?

modules/mysql/mysql.go
```diff
@@ -2,9 +2,7 @@ package mysql
 
 import (
 	"database/sql"
-	"fmt"
-	"strconv"
-	"strings"
+	"sync"
 	"time"
 
 	_ "github.com/go-sql-driver/mysql"
```
prologic, line 8: Is this blank import needed?

modules/mysql/collect.go
```diff
@@ -0,0 +1,103 @@
...
+			m.Errorf("error on collecting user statistics: %v", err)
+			m.doUserStatistics = false
+		}
+	}
+
+	calcThreadCacheMisses(collected)
+	return collected, nil
+}
+
+func (m MySQL) isMariaDB() bool {
```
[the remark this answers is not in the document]
ilyam8: No, it shouldnt

modules/mysql/mysql_test.go
```diff
@@ -0,0 +1,759 @@
...
+	"bufio"
+	"bytes"
+	"database/sql/driver"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
```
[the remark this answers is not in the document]
ilyam8: 🐕

modules/mysql/mysql.go
ilyam8: yes

> https://github.com/go-sql-driver/mysql#usage

modules/mysql/mycnf.go
ilyam8: dunno, i found 

> https://github.com/go-sql-driver/mysql/issues/542

there is this comment in OP

> great job at this is Python's PyMySQL

i decided to check Python's PyMySQL code and implemented something like it does

I saw there are some libraries but all of them not popular (based on stars). Since the thing is not diffucult and basically is just reading/parsing the config in `ini` format i decided to write the code myself.

prologic: Fair

modules/mysql/mysql.go
prologic: Right

prologic:
Still LGTM

modules/mysql/collect_slave_status.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return strconv.ParseInt(value, 10, 64)
+}
+
+func convertSlaveSQLRunning(value string) string {
+	switch value {
+	case "Yes":
+		return "1"
+	default:
+		return "0"
```
Daniel15, line 92: This should probably be `-1` for parity with the Python version? This change broke the alerts. https://github.com/netdata/netdata/issues/10427

PR #406, opened 2020-07-27
[unbound] add expired cache data

and0x000:
add parsing for the `num.expired` statistic.

See: https://nlnetlabs.nl/documentation/unbound/unbound-control/


and0x000:
`Go` is not my native language. So this is mostly just "copy/paste/replace". If I missed some conventions, please, by all means, correct me.

ilyam8:
i wonder did they replace `threadX.num.zero_ttl` with `threadX.num.expired`?

I see no `zero_ttl` in the https://nlnetlabs.nl/documentation/unbound/unbound-control/

I see no `expired` in the https://manpages.debian.org/testing/unbound/unbound-control.8.en.html


```
threadX.num.zero_ttl
  number of replies with ttl zero, because they served an expired cache entry.
```

```
threadX.num.expired
  number of replies that served an expired cache entry.
```

ilyam8:
Yeah, that is the case

https://github.com/NLnetLabs/unbound/blob/69f0adf83b21926bdae85c3b474f83d7ee225fba/doc/Changelog#L391

and0x000:
Thank you for pointing this out.

What is your policy on backward compatibility?

The change on unbound is rather fresh. It is the stable version of unbound, but there are older versions out in the wild.

Debian Buster (current stable) e.g. provides unbound 1.9. Current Ubuntu LTS does so as well.
See: 
* https://packages.debian.org/buster/unbound
* https://packages.ubuntu.com/focal/unbound

If you decide to drop the support of the `num.zero_ttl` statistic for older unbound versions, I can remove it.

Ferroin:
@ilyam8 Just my 2 cents, but I think we should probably try to support both, but I don't know that we need to go to the proably significant effort to provide data continuity across an upgrade of Unbound here.

ilyam8:
Yeah, we need to support both. We need to think how to achieve it.

and0x000:
Any news on that topic? Is there probably a concept for ignoring metrics, if they cannot be found for a service, similar to how services are not shown, when they are not discovered?

ilyam8:
@and0x000 i think we could second chart, i dont like it but it is acceptable, we can fix it later. What i suggest is to change chart titles and units. Both `zero_ttl_replies` and `cache_expired` should have same titles and units.

modules/unbound/charts.go
```diff
@@ -182,6 +185,17 @@ var (
 			{ID: "total.num.prefetch", Name: "prefetches"},
 		},
 	}
+	expiredChart = Chart{
+		ID:       "cache_expired",
+		Title:    "Cache Expiries",
```
ilyam8, line 190: lets make Title `Replies Served From Expired Cache` and units `replies`

ilyam8 (changes requested):

modules/unbound/charts.go
and0x000: requested changes applied with c713836727bea7665e419d878e25a3903805fc6d

ilyam8 (approved):

ilyam8:
Thx @and0x000 !

PR #413, opened 2020-08-17
Add content and clean up Prometheus doc

joelhans:
Nothing major, just doing a little work as I familiarize myself with the Prometheus collector and scope out whether we should do some bigger content on it.


ilyam8:
thx!

PR #417, opened 2020-08-20
Added JSON log parser

mklepaczewski:
This is a JSON log parser. It reads JSON files and maps fields according to the `json_config->mapping` configuration. The parser expects each line to contain single JSON entry.

If this PR is accepted, I'll push small weblog PR that makes use of JSON parser.

# Why

JSON logs are more verbose than CSV-ish files, but JSON has some clear advantages:
- it's much easier to parse and query JSON files from command line(for example using `jq`) for quick troubleshooting,
- it's easier to import them into other tools/services for further processing,
- data types might be preserved.

Some tools provide option to generate JSON logs, e.g. nginx.

# How

`json.Unmarshall()` is being used to decode JSON files. All values read by `json.Unmarshall()` are converted back to strings before they're passed to `LogLine`.

# Todo, future optimization

- ~~JSON file autodetection~~ (done, #1e673b8480050f17943437f8b3becfd08d86b9e5),
- if needed, JSON parsing might be optimized as LogLine accepts only strings and `json.Unmarshall()` type conversion is unnecessary.

### DIFF
--- a/pkg/logs/json.go
+++ b/pkg/logs/json.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"unsafe"
 
 	"github.com/valyala/fastjson"
 )
@@ -57,7 +56,7 @@ func (p *JSONParser) Parse(row []byte, line LogLine) error {
 			return
 		}
 
-		name := *(*string)(unsafe.Pointer(&key))
+		name := string(key)
 		if mapped, ok := p.mapping[name]; ok {
 			name = mapped
 		}
--- a/pkg/logs/json.go
+++ b/pkg/logs/json.go
@@ -61,7 +61,7 @@ func (p *JSONParser) Parse(row []byte, line LogLine) error {
 }
 
 func (p *JSONParser) mapField(field string) string {
-	if newLogLineField,exist := p.mapping[field]; exist == true {
+	if newLogLineField,exist := p.mapping[field]; exist {
 		return newLogLineField
 	}
 	return field
--- a/pkg/logs/json.go
+++ b/pkg/logs/json.go
@@ -12,7 +12,7 @@ type JSONConfig struct {
 }
 
 type JSONParser struct {
-	reader *bufio.Reader
+	reader  *bufio.Reader
 	mapping map[string]string
 }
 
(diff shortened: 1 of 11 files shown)


mklepaczewski:
Please note that this is the first go script I ever wrote - please spend extra minute when reviewing the code ;-)

ilyam8:
Good stuff @mklepaczewski !

For performance, i think, we could use not std json lib like:

 - https://github.com/valyala/fastjson
 - https://github.com/mailru/easyjson

But it can be done in another PR since it is implementation details.

---

Some comments:
 - there are some `reviewdog` warnings that should be fixes
 - `go fmt` the code
 - [weblog readme](https://github.com/netdata/go.d.plugin/tree/master/modules/weblog#log-parsers) should be updated

mklepaczewski:
@ilyam8 Thanks. I think I fixed all issues. I planned to add JSON parser to the `weblog` module in a separate PR, but I guess we can do it here as well. You mentioned that `reviewdog` raised some warnings; however, I've seen & fixed only one - maybe I missed something as go is an entirely new environment to me.

mklepaczewski:
I think all code change requests have been done.

pkg/logs/json.go
```diff
@@ -0,0 +1,72 @@
...
+	var parsedLine map[string]interface{}
+
+	if err := json.Unmarshal(row, &parsedLine); err != nil {
+		return err
+	}
+
+	/* Now map the fields */
+	for logField, logValue := range parsedLine {
+		/* Convert logValue to string */
+		stringValue := fmt.Sprintf("%v", logValue)
```
ilyam8, line 53: Why do we need this? Are there something we cant unmarshal into a `string`?

mklepaczewski: I convert it back to string because `weblog/logline::Assign()` expects the `value` argument to be a string. The json log that is being parsed might store integers either as strings or, well, integers. In effect we're not sure of the type of the value, hence the cast to string. The same goes for floats and bools.

ilyam8:
> @mklepaczewski i think we need to switch to valyala/fastjson

if you want i can do it

mklepaczewski:
@ilyam8 I'm sorry, I've been quite busy. Please do change it if it's not a problem. As I haven't worked with `fastjson` before (again, I'm completely to Go in general) this is a non-trivial change for me, even if the change itself is trivial ("unknown unknowns")

ilyam8 (approved):

ilyam8:
@mklepaczewski [i removed your config](https://github.com/netdata/go.d.plugin/pull/417/commits/dabf289d888565fd485c281d38f297cf781a33af) because it doesnt look generic, so there is no point to try this job.

ilyam8:
@mklepaczewski thx for contributing, json parser is a nice thing to have!

Issue #419, opened 2020-08-26
Improve README.md for matcher pkg


OdysLam:
A user contacted support to setup the collector for openvpn. The issue was an incorrect understanding of the matcher syntax. The agent (me) also failed to understand how the pattern syntax works. The issue was resolved with the help from @ilyam8.

Thus, we need to improve the docs on the matcher in order to make it easier and simpler for people to use the patterns (which are already simpler than regex).  Some examples will help the user to understand the syntax and use this feature successfully.

## Resources
Forum Post: https://community.netdata.cloud/topic/65/how-to-use-properly-openvpn-conf-module-so-we-can-have-per_user_stats/2
Slack discussion: https://netdata-cloud.slack.com/archives/CSD2D3NTG/p1598438425053500

ilyam8:
fixed in #422

PR #422, opened 2020-08-28
Add examples to readme.md

OdysLam:


pkg/matcher/README.md
```diff
@@ -4,12 +4,18 @@
   * glob
   * regexp
   * simple patterns
+
+Depending on the symbol at the start of the string, the `matcher` will use one of the supported formats. 
+
+e.g `'* *'`: Because of the first `*`, `matcher` understands that it's in `glob` format, and according to the `glob` format the second `*` means that it will match *any* string.
```
ilyam8, line 10: I think this **specific** `e.g` shouldnt be there, at this point user is not aware of syntax so glob/not glob doesnt say much.

> Depending on the symbol at the start of the string, the `matcher` will use one of the supported formats

This is good because it doesnt contain specific details.

What we could do it to add a table like


| matcher           | short format  | long format           |
|-------------------|---------------|-----------------------|
| string            | `=`           |    `string`           |
| glob              | `*`           |    `glob`             |
| regexp            | `~`           |    `regexp`           |
| simple patterns   |               |    `simple_patterns`  |

pkg/matcher/README.md
```diff
@@ -21,19 +27,30 @@ Short Syntax
 
  Long Syntax
 
-     <line>      ::= [ <not> ] <format> <separator> <expr>
+     [ <not> ] <format> <separator> <expr>
      <format>    ::= [ 'string' | 'glob' | 'regexp' | 'simple_patterns' ]
```
ilyam8, line 31: remove a new line after `Long Syntax` and add after `[ <not> ] <format> <separator> <expr>`

pkg/matcher/README.md
```diff
@@ -21,19 +27,30 @@ Short Syntax
...
      <not>       ::= '!'
                        negative expression
      <separator> ::= ':'
      <expr>      ::= any string
 ```
 
+**Tip**: In short syntax, you can enable the glob format by starting the string with a `*`, while in the long syntax you need to define it more explicitly. The following examples are identical. `simple_patterns` can be used **only** with the long syntax.
+
+Short Syntax: `'* * '`
+Long Syntax: `'glob':'*'`
```
ilyam8, line 41: i think we dont need this tip, all specific examples should be under appropriate matchers sections

pkg/matcher/README.md
```diff
@@ -21,19 +27,30 @@ Short Syntax
...
+
 ### String matcher
 The string matcher reports whether the given value equals to the string ( use == ).
```
ilyam8, line 44: `( use == )` => `( use = )`

pkg/matcher/README.md
```diff
@@ -21,19 +27,30 @@ Short Syntax
...
 
+e.g <br>
+`'== foo'` will match only if the string is `'foo'`. <br>
+`'[!] == bar'` will match any string that is not `'bar'`.
```
ilyam8, line 48: i think the following would be better

```
Examples:
-   <EXAMPLE_1>
-   <EXAMPLE_2>
```
---

`[ ]` means **optional** (see any `man` page), so it is `!= PATTERN`

pkg/matcher/README.md
```diff
@@ -52,16 +69,30 @@ The pattern syntax is:
         '\\' c      matches character c
         lo '-' hi   matches character c for lo <= c <= hi
 ```
+e.g 
+ - `* ?` will match any string that is a single character.
```
ilyam8, line 73:
```
Examples:
-   <EXAMPLE_1>
-   <EXAMPLE_2>
```

pkg/matcher/README.md
```diff
@@ -52,16 +69,30 @@ The pattern syntax is:
...
+ - `'*[a-d]'` will match any string (`*`) that ends with a character that is between `a` and `d` (i.e `a,b,c,d`).
+
+
       
 ### Regexp matcher
 The regexp matcher reports whether the given value matches the RegExp pattern ( use regexp.Match ).
 
 The RegExp syntax is described at https://golang.org/pkg/regexp/syntax/.
 
+We know that Regular Expressions are hard, but they are powerfull. If you want to learn more about them, you can visit [RegexOne](https://regexone.com/).
```
ilyam8, line 85: https://regex101.com/ is awesome

again


```
Examples:
-   <EXAMPLE_1>
-   <EXAMPLE_2>
```

pkg/matcher/README.md
```diff
@@ -52,16 +69,30 @@ The pattern syntax is:
...
+
 ### Simple patterns matcher
 The simple patterns matcher reports whether the given value matches the simple patterns.
 
-The simple patterns is a custom format used in netdata, it's syntax is described at https://docs.netdata.cloud/libnetdata/simple_pattern/.
+Simple patterns are a space separated list of words, that can have `*` as a wildcard. Each world may use any number of `*`. Simple patterns allow negative matches by prefixing a word with `!`.
+
+So, pattern = `!*bad* *` will match anything, except all those that contain the word bad.
+
+Simple patterns are quite powerful: pattern = `*foobar* !foo* !*bar *` matches everything containing foobar, except strings that start with foo or end with bar.
```
ilyam8, line 94: This copied from simple patterns readme, i think we shouldnt do it and just provide a link. Some example would be good tho!

pkg/matcher/README.md
ilyam8: could be

| format          | description      |
|-------------------|---------------|
| `=`            | Match values that are exactly equal to the provided string           |
| `!=`            | Match values that are not equal to the provided string           |
| `*`            | Match values that glob-match the provided string           |
| `!*`            | Match values that do not glob-match the provided string           |
| `~`            | Match values that regexp-match the provided string           |
| `!~`            | Match values that do not regexp-match the provided string           |

ilyam8:
@joelhans cant assign you as a reviewver, seems like some repo permissions 🤷‍♂️  But please!

OdysLam:
This is tracked by issue #419

pkg/matcher/README.md
OdysLam: I think we should provide all the necessary information in the readme, instead of redirecting them to our docs, we add useless friction.

pkg/matcher/README.md
OdysLam: I want this example because it makes it explicit that the 2 formats are identical. it took me some time to understand what short and long format is, this example makes it dead obvious, so I would like to keep it.

pkg/matcher/README.md
ilyam8: I dont think it is useless, the up-to-date description is netdata simple-patterns readme. Same goes for regex - we dont want to describe regex syntax when there is official page with the syntax

pkg/matcher/README.md
```diff
@@ -4,12 +4,25 @@
...
-  
+
+Depending on the symbol at the start of the string, the `matcher` will use one of the supported formats. 
+
+matcher | short format | long format
+ ------ | ------------ | ----------
+ string | ` =` | `string`
+ glob | `*` | `glob`
+ regexp | `~` | `regexp`
+ simple patterns | | `simple patterns`
```
ilyam8, line 15: `simple patterns` => `simple_patterns`

pkg/matcher/README.md
ilyam8: you added the table in the beginning of readme, that info in there

pkg/matcher/README.md
```diff
@@ -20,20 +33,31 @@ Short Syntax
...
+Short Syntax: `'* * '`
+Long Syntax: `'glob':'*'`
+
 ### String matcher
-The string matcher reports whether the given value equals to the string ( use == ).
+The string matcher reports whether the given value equals to the string ( use = ).
+
+Examples: 
+ - `'== foo'` will match only if the string is `'foo'`.
+ - `'[!] == bar'` will match any string that is not `'bar'`.
```
ilyam8, line 55: it is one `=`

pkg/matcher/README.md
ilyam8: Lets add it as an `Examples` section. Same as you did in `glob` matcher section.

```
Examples:
- `!*bad* *` will match anything, except all those that contain the word bad.
- ...
```

ilyam8: Ok i have a second thought and i think you are right, we dont need the link to the netdata simple_patterns at all because go.d simple patterns has more features (for example netdata doesnt handle `[a-z]`).

So lets get Examples section here and it is ok

ilyam8: simple patterns is basically a space separated list of glob patterns. Since it is a list we try them in turn and the result is the first one that mathes. In addition a glob pattern can be negated (means if it mathces it returns `false`)

pkg/matcher/README.md
ilyam8: Here we could add some tips.

String matcher means exact match. There are other string match related cases:

- has prefix
- has suffic
- contains

As you can guess it is achievable using `glob` matcher

- has prefix: `PREFIX*`
- has suffic: `*SUFFIX`
- contains: `*SUBSTRING*`

But it could be worth to mention in this section because these are kind of string matching cases.

What do you think?

pkg/matcher/README.md
ilyam8: i would add negative matchers here too

ilyam8: and a description, like i have it there

> https://github.com/netdata/go.d.plugin/tree/master/pkg/prometheus/selector#syntax-1

but table is better then list in this case, yes

pkg/matcher/README.md
OdysLam: Still, I prefer to over-communicate things that people may have already understand, that miss something that a few people will scratch their heads about. I feel quite strongly about this one tbh :P @joelhans wanna chime in ? 
In general my approach to documentation is "assume that you are talking to a small child, or a even a dog".

pkg/matcher/README.md
```diff
@@ -42,18 +44,30 @@ Short Syntax
...
 ```
 
-**Tip**: In short syntax, for example, you can enable the glob format by starting the string with a `*`, while in the long syntax you need to define it more explicitly. The following examples are identical. `simple_patterns` can be used **only** with the long syntax.
+**Remember**: In short syntax, for example, you can enable the glob format by starting the string with a `*`, while in the long syntax you need to define it more explicitly. The following examples are identical. `simple_patterns` can be used **only** with the long syntax.
 
-Short Syntax: `'* * '`
-Long Syntax: `'glob':'*'`
+Examples:
+ - Short Syntax: `'* * '`
+ - Long Syntax: `'glob':'*'`
```
ilyam8, line 51: `'glob':'*'` => `'glob:*'`

pkg/matcher/README.md
```diff
@@ -42,18 +44,30 @@ Short Syntax
...
 
 ### String matcher
 The string matcher reports whether the given value equals to the string ( use = ).
 
 Examples: 
- - `'== foo'` will match only if the string is `'foo'`.
- - `'[!] == bar'` will match any string that is not `'bar'`.
+ - `'= foo'` will match only if the string is `'foo'`.
+ - `'! = bar'` will match any string that is not `'bar'`.
```
ilyam8, line 58: `! = bar` => `!= bar`

ilyam8: is `'foo' => is `foo`, same for `bar`

pkg/matcher/README.md
```diff
@@ -42,18 +44,30 @@ Short Syntax
...
+
+String matcher means **exact match** of the `string` with the `pattern`. There are other string match related cases:
```
ilyam8, line 60: > String matcher means **exact match** of the `string` with the `pattern`

in that case it is not `pattern` but an exact value

pkg/matcher/README.md
```diff
@@ -20,20 +35,43 @@ Short Syntax
...
+     [ <not> ] <format> <separator> <expr>
+     
      <format>    ::= [ 'string' | 'glob' | 'regexp' | 'simple_patterns' ]
      <not>       ::= '!'
                        negative expression
      <separator> ::= ':'
      <expr>      ::= any string
 ```
 
+**Remember**: In short syntax, for example, you can enable the glob format by starting the string with a `*`, while in the long syntax you need to define it more explicitly. The following examples are identical. `simple_patterns` can be used **only** with the long syntax.
```
joelhans, line 47:
```suggestion
When using the short syntax, you can enable the glob format by starting the string with a `*`, while in the long syntax you need to define it more explicitly. The following examples are identical. `simple_patterns` can be used **only** with the long syntax.
```

pkg/matcher/README.md
```diff
@@ -20,20 +35,43 @@ Short Syntax
...
+
+Examples:
+ - Short Syntax: `'* * '`
+ - Long Syntax: `'glob':'*'`
+
 ### String matcher
-The string matcher reports whether the given value equals to the string ( use == ).
+The string matcher reports whether the given value equals to the string ( use = ).
```
joelhans, line 54: I'm not sure if the "( use = )" text is necessary here. The examples below explain the syntax, as does the table at the beginning.

pkg/matcher/README.md
joelhans: I would agree that this is confusing if we're saying the string matcher looks for exact matches, then also say it's a pattern. 

I actually would recommend we move this glob content down into the next section about globs as some additional examples of how to capture strings using globs.

pkg/matcher/README.md
```diff
@@ -20,20 +35,43 @@ Short Syntax
...
+This is achievable using the `glob` matcher:
+
+ - `* PREFIX*`, means that it will match with any string that *starts* with `PREFIX`, e.g `PREFIXnetdata`
+ - `* *SUFFIX`, means that it will match with any string that *ends* with `SUFFIX`, e.g `netdataSUFFIX`
+ - `* *SUBSTRING*`, means that it will match with any string that *contains* `SUBSTRING`, e.g `netdataSUBSTRINGnetdata`
 
 ### Glob matcher
-The glob matcher reports whether the given value matches the wildcard pattern.
+
+The glob matcher reports whether the given value matches the wildcard pattern. It uses the standard `golang` library `path`. You can read more about the library in the [golang documentation](https://golang.org/pkg/path/#Match), where you can also practice with the library in order to learn the syntax and use it in your netdata configuration.
```
joelhans, line 74:
```suggestion
The glob matcher reports whether the given value matches the wildcard pattern. It uses the standard `golang` library `path`. You can read more about the library in the [golang documentation](https://golang.org/pkg/path/#Match), where you can also practice with the library in order to learn the syntax and use it in your Netdata configuration.
```

pkg/matcher/README.md
```diff
@@ -52,16 +90,30 @@ The pattern syntax is:
...
+ - `'*[a-d]'` will match any string (`*`) that ends with a character that is between `a` and `d` (i.e `a,b,c,d`).
+
+
       
 ### Regexp matcher
 The regexp matcher reports whether the given value matches the RegExp pattern ( use regexp.Match ).
 
 The RegExp syntax is described at https://golang.org/pkg/regexp/syntax/.
 
+We know that Regular Expressions are hard, but they are powerfull. If you want to learn more about them, you can visit [RegexOne](https://regexone.com/).
```
joelhans, line 106:
```suggestion
Learn more about regular expressions at [RegexOne](https://regexone.com/).
```

pkg/matcher/README.md
```diff
@@ -52,16 +90,30 @@ The pattern syntax is:
...
+
 ### Simple patterns matcher
 The simple patterns matcher reports whether the given value matches the simple patterns.
 
-The simple patterns is a custom format used in netdata, it's syntax is described at https://docs.netdata.cloud/libnetdata/simple_pattern/.
+Simple patterns are a space separated list of words, that can have `*` as a wildcard. Each world may use any number of `*`. Simple patterns allow negative matches by prefixing a word with `!`.
```
joelhans, line 111:
```suggestion
Simple patterns are a space separated list of words. Each word may use any number of wildcards `*`. Simple patterns allow negative matches by prefixing a word with `!`.
```

joelhans (changes requested):
A few thoughts for now as I'm continuing to wrap my head around this stuff. Once we hammer out all these details let's push this so it's out before the 1.25 release.

Next week I'm prioritizing setting up some systems/configs that require this syntax so I can better understand how it works, and then we can assess whether more or different content is necessary.

pkg/matcher/README.md
joelhans: I think we should at least link to the simple patterns doc as a reference to that source of truth, even if they are slightly different implementations. Overall I think we need to increase visibility on the simple patterns syntax, as it's something that I've seen a few users stumble over since it's not a "standard" like regex.

pkg/matcher/README.md
ilyam8: ^^ this mainly a note for me 😄

pkg/matcher/README.md
ilyam8: It is not really source of truth. It behaves same as netdata simple patterns in general:

- list of patterns
- execute them in turn
- first one that matches returns

pattern syntax itself is different, in addition to `*` go.d simple patterns support `?` `[...]` `[^...]

That is why i started to think that we dont need it 🤔

OdysLam:
Implemented the changes you requested @ilyam8 @joelhans

ilyam8 (approved):

joelhans (approved):

ilyam8:
thx @OdysLam !

PR #439, opened 2020-09-17
add Systemd unit states collector

hamedbrd:
This is a new module for monitoring systemd units, before adding the documentation and tests I need to check the functionality and code style, once we agreed on these then I'm going to add documentation and of course some tests to it.

Fixes: netdata/netdata#6683


modules/systemdstates/collect.go
```diff
@@ -0,0 +1,97 @@
...
+			state = inactive
+		}
+		if unit.ActiveState == "failed" {
+			state = failed
+		}
+		mx[unit.Name] = int64(state)
+	}
+
+	return mx, nil
+}
```
ilyam8, line 49: We need:

- collect units
- remove units if
  - `LoadState` != `loaded`
  - selector doesnt match
- determine unit type (service, socket, etc.)
- check if an appropriate chart (we have a chart per type) has dimension
  - has no: add dimension
  - has: do nothing
- convert string status to a numeric value and populate map[string]int64

ilyam8 (changes requested):

modules/systemdstates/systemdstates.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Check makes check
+func (SystemdStates) Check() bool {
+	return true
+}
+
+// Charts creates Charts
+func (s *SystemdStates) Charts() *Charts {
+	return s.charts()
+}
```
ilyam8, line 63: we need to add `charts` fields to the main structure and return it here

that is what i mean:

```
func New() *SystemdStates {
	return &SystemdStates{
		metrics: make(map[string]int64),
        charts: charts.Copy(),
	}
}

...

type SystemdStates struct {
	module.Base  // should be embedded by every module
	Config       `yaml:",inline"`
	metrics      map[string]int64
	unitsMatcher matcher.Matcher
    charts       *module.Charts
}

func (s *SystemdStates) Charts() *Charts {
	return s.charts
}
```

ilyam8: Reason: this is always the pattern if you need to change charts (add/remove charts, add/remove dimensions) in runtime, you add charts instance as an private field to the main structure to get access to it later.

`Charts()` must return instance specific charts, so if there are any changes plugin can detect it and resend charts definition to netadta daemon.

modules/systemdstates/collect.go
hamedbrd: please take a look again and let me know what more is needed

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,89 @@
...
+		if unit.ActiveState == "active" {
+			state = active
+		}
+		if unit.ActiveState == "inactive" {
+			state = inactive
+		}
+		if unit.ActiveState == "failed" {
+			state = failed
+		}
+		mx[unit.Name] = int64(state)
```
ilyam8, line 56: lets create `convertUnitState(state string) int` functions

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,89 @@
...
+		if strings.HasSuffix(unit, fmt.Sprintf(".%s", t)) {
+			ut = t
+			break
+		}
+	}
+
+	if ut == "" {
+		return "", fmt.Errorf("Could not find a type for : %v", unit)
+	}
+	return ut, nil
```
ilyam8, line 88: it is better to make it a function

`func extractUnitType(name string) (string, error)`

and use `switch` in there

ilyam8 (changes requested):

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,89 @@
...
+	filtered := make([]dbus.UnitStatus, 0, len(units))
+	for _, unit := range units {
+
+		if s.unitsMatcher.MatchString(unit.Name) && unit.LoadState == "loaded" {
+			filtered = append(filtered, unit)
+		}
+	}
+
+	return filtered
+}
```
ilyam8, line 73: i suggest to use this approach for filtering, we can use same slice

https://github.com/netdata/go.d.plugin/blob/b1e849a145abe8e620e49fb9b884ce70ef564d6e/modules/elasticsearch/collect.go#L286-L296

> s.unitsMatcher.MatchString(unit.Name) && unit.LoadState == "loaded"

i think `unit.LoadState == "loaded"` should be first

ilyam8:
and please do `go mod tidy`, there could be some leftovers from the old systemd lib

modules/systemdstates/collect.go
hamedbrd: it's a function, do you mean method is not needed and also using a switch instead of for?

ilyam8: yeah, that is what i mean, we dont need access to systemd structure to extract a unit type

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,90 @@
...
+		return 1
+	case "inactive":
+		return 0
+	case "failed":
+		return 2
+	default:
+		return -1
+	}
+
+}
```
ilyam8, line 23: isn't there more states?

> []string{"active", "activating", "deactivating", "inactive", "failed"}

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,90 @@
...
+		if strings.HasSuffix(unit, fmt.Sprintf(".%s", t)) {
+			ut = t
+			break
+		}
+	}
+
+	if ut == "" {
+		return "", fmt.Errorf("Could not find a type for : %v", unit)
+	}
+	return ut, nil
```
ilyam8, line 89: use `switch` reminder

we know that unit name pattern is *.unitType

use `strings.LastIndexByte`

ilyam8: I am not sure we need to check if type is valid or not.

I see current implementation expects only the following types

```go
	validTypes := []string{"service", "socket", "device", "mount", "automount", "swap", "target", "path", "timer", "scope"}
```

because we have predefined charts.

---

so it is ok to validate but lets make it a sep function `isUnitTypeValid`

`extractUnitType`:
 - extract
 - check if it is valid
 - return type, err

hamedbrd: ok, I changed it with a switch, and also LastIndexByte to extract unit name

modules/systemdstates/collect.go
hamedbrd: all of these types are added now

modules/systemdstates/collect.go
```diff
@@ -74,16 +78,33 @@ func (s SystemdStates) filterUnits(units []dbus.UnitStatus) []dbus.UnitStatus {
...
-	validTypes := []string{"service", "socket", "device", "mount", "automount", "swap", "target", "path", "timer", "scope"}
 	ut := ""
-	for _, t := range validTypes {
-		if strings.HasSuffix(unit, fmt.Sprintf(".%s", t)) {
-			ut = t
-			break
-		}
+	idx := strings.LastIndexByte(unit, '.')
+
+	switch suffix := unit[idx:]; suffix {
```
ilyam8, line 84: i suggest

```go
idx := strings.LastIndexByte(unit, '.')
if idx <= 0 {
	...
}
unitType := unit[idx+1:]
```

^^ that is extract already and we can return w/o switch.

I see you want to add some type validation which is fine by me.

I suggest to make it a separate function and use in `extractUnitType` 

```go
idx := strings.LastIndexByte(unit, '.')
if idx <= 0 {
	...
}
unitType := unit[idx+1:]
if !isUnitTypeValid(unitType) {
   ...
}
...
```

modules/systemdstates/systemdstates.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+// SystemdStates SystemdStates module
+type SystemdStates struct {
+	module.Base  // should be embedded by every module
+	Config       `yaml:",inline"`
+	metrics      map[string]int64
+	charts       *module.Charts
+	unitsMatcher matcher.Matcher
```
ilyam8, line 38: Lets call it `selector` or (`sr`). Configuration option has `selector` name too.

`metrics      map[string]int64` field is not needed (it is not used)

modules/systemdstates/systemdstates.go
```diff
@@ -0,0 +1,79 @@
...
+		s.unitsMatcher = matcher.WithCache(m)
+	}
+
+	return true
+}
+
+// Check makes check
+func (SystemdStates) Check() bool {
+	return true
+}
```
ilyam8, line 61: I think this should be simple `return len(s.Collect()) > 0`.

We use check to ensure the module is able to collect data (plugin invokes Check() before starting the module).

modules/systemdstates/charts.go
```diff
@@ -0,0 +1,93 @@
...
+	Charts = module.Charts
+
+	// Dim is an alias for module.Dim
+	Dim = module.Dim
+)
+
+var charts = Charts{
+	{
+		ID:    "service_states",
+		Title: "Systemd Service Units (active => 1, activating => 2, failed => 3, inactive => 4, deactivating => 5)",
```
ilyam8, line 18: i think we need to do it other way arount: `num => text`, it looks more logical to me because we decode it that way - numeric representation to textual.

And i suggest to use `num: text, ` pattern, it is used it in several collectors

ilyam8: (for all charts ofc)

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/coreos/go-systemd/v22/dbus"
+)
+
+func convertUnitState(state string) int64 {
+
```
ilyam8, line 11: an empty line (11) is not needed

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,101 @@
...
+	case "deactivating":
+		return 5
+	default:
+		return -1
+	}
+
+}
+
+func (s *SystemdStates) collect() (map[string]int64, error) {
+
```
ilyam8, line 30: empty line(30)

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,101 @@
...
+	var err error
```
ilyam8, line 31: `var err error` is not needed

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,101 @@
...
+	allUnits, err := conn.ListUnits()
+	if err != nil {
+		return nil, err
+	}
+
+	mx := make(map[string]int64)
+
+	units := s.filterUnits(allUnits)
+	for _, unit := range units {
+
```
ilyam8, line 47: empty line (47)

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+
+		mx[unit.Name] = convertUnitState(unit.ActiveState)
+	}
+
+	return mx, nil
+}
+
+func (s SystemdStates) filterUnits(units []dbus.UnitStatus) []dbus.UnitStatus {
+
```
ilyam8, line 66: empty line(66)

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, unit := range units {
+
+		if unit.LoadState == "loaded" && s.selector.MatchString(unit.Name) {
+			units[i] = unit
+			i++
+		}
+	}
+
+	return units[:i]
+
```
ilyam8, line 77: empty line(77)

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func extractUnitType(unit string) (string, error) {
+
```
ilyam8, line 81: empty line (81)

modules/systemdstates/collect.go
```diff
@@ -0,0 +1,101 @@
...
+	idx := strings.LastIndexByte(unit, '.')
+
+	if idx <= 0 {
+		return "", fmt.Errorf("Could not find a type for : %v", unit)
```
ilyam8, line 85: please start all errors with an lowercase letter

modules/systemdstates/systemdstates.go
```diff
@@ -0,0 +1,77 @@
...
+	charts      *module.Charts
+	selector    matcher.Matcher
+}
+
+// Cleanup makes cleanup
+func (SystemdStates) Cleanup() {}
+
+// Init makes initialization
+func (s *SystemdStates) Init() bool {
+
```
ilyam8, line 44: empty line(44)

ilyam8:
@hamedbrd i think we need to name it `systemdunits`

modules/systemdunits/collect.go
```diff
@@ -0,0 +1,94 @@
...
+		ut, err := extractUnitType(unit.Name)
+		if err != nil {
+			return nil, err
+		}
+
+		chartID := fmt.Sprintf("%s_states", ut)
+		chart := s.charts.Get(chartID)
+		if !chart.HasDim(unit.Name) {
+			_ = chart.AddDim(&Dim{ID: unit.Name})
+		}
```
ilyam8, line 53: `charts` and `dimensions` are slices, `Get` and `HasDim` methods are not fast - it is iterating over the slice.

This is a problem and should be fixed, but for now you need to keep track of `collected` services in the module.

See example:
https://github.com/netdata/go.d.plugin/blob/79ffe8973780b29f1edae13550e0070fdfb69c37/modules/filecheck/collect_files.go#L18-L22

hamedbrd: I got this idea from other modules, like this one
https://github.com/netdata/go.d.plugin/blob/79ffe8973780b29f1edae13550e0070fdfb69c37/modules/k8s_kubelet/collect.go#L46-L49

ilyam8: i see, it is not good, imagine iterating over the same slice on every iteration number of units times again and again.

ilyam8:
We dont want to test extractUnite and filtering, those should be in the `TestCollect` test cases.

modules/systemdunits/systemdunits.go
```diff
@@ -0,0 +1,86 @@
...
+func (s *SystemdUnits) Init() bool {
+	if !s.Selector.Empty() {
+		m, err := s.Selector.Parse()
+		if err != nil {
+			s.Errorf("error on creating per user stats matcher : %v", err)
+		}
+		s.selector = matcher.WithCache(m)
+	}
+
+	allUnits, err := s.getUnits()
```
ilyam8, line 56: why `getUnits()` in Init? We call collector's Init functuin only once. `getUnits` should be in collect

ilyam8:
@hamedbrd please rebase on the current master and fix the merge conflicts

modules/systemdunits/systemdunits.go
hamedbrd: that's where I need your advice, I did this because I want to keep `dbus` connections outside `collect` method due to having some fake data while we are running tests. any idea/sample how to handle this?

ilyam8:
Some notes:
- gathering all systemd units is CPU-intesive task (up to 3% cpu usage according apps.plugin), so default `update_every` is 10.
- `go-systemd/dbus` package has [`ListUnitsByPatterns(states []string, patterns []string)`](https://pkg.go.dev/github.com/coreos/go-systemd/v22@v22.1.0/dbus#Conn.ListUnitsByPatterns) method. I compared `ListUnits + filtering after` vs `ListUnitsByPatterns` execution time and 'ListUnitsByPatterns' is quite faster. I think we can use it.
- It seems there is no need to have this collector enabled by default, so disabled by default.

ilyam8 (approved):

thiagoftsm (approved):
LGTM!

config/go.d.conf
```diff
@@ -51,6 +52,7 @@ modules:
 #  solr: yes
 #  springboot2: yes
 #  squidlog: yes
+#  systemdunits: yes
```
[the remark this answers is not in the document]
ilyam8: it should be yes, if you want to enable it you need to uncomment the line

vlvkobal (approved):

ilyam8:
thx @hamedbrd !

PR #444, opened 2020-09-21
Move library between diretories

### LINKED ISSUE #443: move `modules/dnsmasq_dhcp/ip` => `pkg`
We need to move [`modules/dnsmasq_dhcp/ip`](https://github.com/netdata/go.d.plugin/tree/master/modules/dnsmasq_dhcp/ip) pkg to [`pkg`](https://github.com/netdata/go.d.plugin/tree/master/pkg).

We will use it in another collector.

Just move it for now w/o any modification in the pkg.

thiagoftsm:
##### Summary 

Fixes #443 

This PR is moving the directory `modules/dnsmasq_dhcp/ip/` to  `pkg/` and it is also updating the information inside `dnsmasq_dhcp` directory.

##### Test plan

1 - Clone the PR and checkout on it.
2 - Run : `go test ./...`


modules/dnsmasq_dhcp/dhcp.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/netdata/go-orchestrator/module"
 
-	"github.com/netdata/go.d.plugin/modules/dnsmasq_dhcp/ip"
+	"github.com/netdata/go.d.plugin/pkg/ip"
```
ilyam8, line 10: I try to follow the order:

```
std lib

local packages (within-repo)

remote packages
```

ilyam8: `github.com/netdata/go-orchestrator/module` is remote atm, `github.com/netdata/go.d.plugin/pkg/ip` - local.

ilyam8 (changes requested):

modules/dnsmasq_dhcp/dhcp.go
thiagoftsm: Thank you for the tip, I fixed the order inside this file and I will move forward with this pattern.

ilyam8 (approved):

PR #445, opened 2020-09-22
add `filecheck` module

ilyam8:
Fixes: netdata/netdata#9794

Filecheck collects `files` and `directories` statistics.

Files statistics:
  - existence
  - time since the last modification
  - size

Directories statistics:
  - existence
  - time since the last modification
  - number of files

---

It produces the following file check charts (chart dimension - filepath):
  - File Existence in `boolean`
  - File Time Since the Last Modification in `seconds`
  - File Size in `bytes`

Dir check charts (chart dimension - dirpath):
  - Dir Existence in `boolean`
  - Dir Time Since the Last Modification in `seconds`
  - Dir Number of Files in `files`

---

Example configuration:

```yaml
jobs:
 - name: files_dirs_example
   files:
     include:
       - '/path/to/file1'
       - '/path/to/file2'
   dirs:
     include:
       - '/path/to/dir1'
       - '/path/to/dir2'
```

---

Limitations of the current version:
 - file/dir path doesn't support any wildcards, should be an absolute path to a file/dir.
 - `stat` call is used to gather metrics on every iteration for every file/dir.

We will improve this module later, the goal is to provide the functionality.


Ferroin:
Two things come to mind that could probably be easily improved here:

* We should document whether the existence check verifies the inode type. IOW, if we are checking for the existence of a directory, do we actually ensure it's a directory before proceeding to the other checks? If so, it should be documented. If not, then we should almost certainly make it behave this way and then document that it behaves this way.
* It would be nice to be able to skip certain checks on a given job. Obviously this won't eliminate the need to call `stat()` in most cases, but it will help avoid people having extra charts they don't actually need.

ilyam8:
@Ferroin thx for suggestions!

> do we actually ensure it's a directory before proceeding to the other checks

yes, same goes for a file

> It would be nice to be able to skip certain checks on a given job.

It is one `stat` call per item (file, dir).

ilyam8:
> but it will help avoid people having extra charts they don't actually need.

If there are any requests we will add this, np. But not in this PR.

modules/filecheck/README.md
```diff
@@ -0,0 +1,104 @@
...
+       - '/path/to/dir1'
+       - '/path/to/dir2'
+```
+
+For all available options, see the Filecheck collector's [configuration
+file](https://github.com/netdata/go.d.plugin/blob/master/config/go.d/filecheck.conf).
+
+## Limitations
+
+-   fil/dir path pattern doesn't support any wildcards
```
thiagoftsm, line 83: I think there is a typo problem here, instead `fil/dir` I imagine the correct would be `file/dir`, because until now I could not see any reference to `fil`.

modules/filecheck/charts.go
```diff
@@ -0,0 +1,63 @@
...
+var (
+	fileCharts = module.Charts{
+		fileExistenceChart.Copy(),
+		fileModTimeAgoChart.Copy(),
+		fileSizeChart.Copy(),
+	}
+
+	fileExistenceChart = module.Chart{
+		ID:    "file_existence",
+		Title: "File Existence (0: doesnt exist, 1: exists)",
```
thiagoftsm, line 14: Title should be `0: doesn't exist`.

modules/filecheck/charts.go
```diff
@@ -0,0 +1,63 @@
...
+var (
+	dirCharts = module.Charts{
+		dirExistenceChart.Copy(),
+		dirModTimeChart.Copy(),
+		dirNumOfFilesChart.Copy(),
+	}
+
+	dirExistenceChart = module.Chart{
+		ID:    "dir_existence",
+		Title: "Dir Existence (0: doesnt exist, 1: exists)",
```
thiagoftsm, line 44: It is also necessary to adjust the title here, `doesn't` instead `doesnt`.

modules/filecheck/charts.go
ilyam8: I didn't add it deliberately. I guess it can break plugins.d parser. Let me check.

ilyam8: because we wrap chart fields in `'`

https://github.com/netdata/go-orchestrator/blob/724cb40df90f21ef5d34af785146b1ecb7e145f5/pkg/netdataapi/api.go#L33-L34

modules/filecheck/README.md
ilyam8: There is invisible `e` 👻

modules/filecheck/charts.go
thiagoftsm: I got it, but I never saw `doesnt` in English literature. @joelhans is it possible to use `doesnt` ? Or should @ilyam8 to change for ` It does not`?

modules/filecheck/charts.go
ilyam8: see above
i am pretty sure plugins.d parser can't handle `'` inside `'`

We send it as 
```
CHART 'filecheck_netdata_log.file_existence' '' 'File Existence (0: doesnt exist, 1: exists)' 'boolean' 'files' 'filecheck.file_existence' '' '70000' '10' '' 'go.d' 'filecheck'
```

modules/filecheck/charts.go
ilyam8: @joelhans is it ok to use `exist/not exist` ?

ilyam8: fixed, sql has [`exists/not exists`](https://vladmihalcea.com/sql-exists/)

modules/filecheck/init.go
```diff
@@ -0,0 +1,35 @@
+package filecheck
+
+import (
+	"errors"
+
+	"github.com/netdata/go-orchestrator/module"
+)
+
```
thiagoftsm, line 8: I want to confirm one conclusion, for the next two `methods`, you are copying the whole object instead to give as argument. Are you doing this because they are only called during the initialization ?

ilyam8: i'm not copying anyting, i'm importing that package

thiagoftsm: Sorry, I did not express well. I was talking about [this situation](https://tour.golang.org/methods/4). The methods does not have pointers, so you are copying the whole data before to do the tests, or am I missing something?

ilyam8: can you show the exact part of the code you are talking about? you highlighted `import`...

ilyam8: ok, i think i get it, you are talking about `Filecheck` methods in the `init.go` file.

A function/method declaration should tell you a story - what it does, what it want, what it produces.

Ptr receiver means you can modify the value to which the receiver points. So if it is so i would assume you do so.

In that situation it is not ptr receiver because i dont need to change anything and you can say that just by the method declaration.

The rule is:
 - if you need to modify the value to which the receiver points - ptr rcvr
 - if you dont - value rcvr
 - if in doubts - ptr rcvr

> copying the whole data before

I suggest to not take it into account atm, there are cases when you want to have ptr rcvr just for optimization, but this case is **far** from it.

Copying is everywhere (it is either value copy or ptr copy), it is ok until it is a lot of heap allocations/ memory leak.

So for now just follow the rules i've written above.

thiagoftsm:
I created the following structure to test:

```
bash-5.0$ ls -lh 
total 4.0K
-rw-r--r-- 3 thiago thiago    0 Sep 22 18:51 hard_link
drwxr-xr-x 2 thiago thiago 4.0K Sep 22 18:51 mydir
-rw-r--r-- 3 thiago thiago    0 Sep 22 18:51 myfile
lrwxrwxrwx 1 thiago thiago    5 Sep 22 18:51 soft_link -> mydir
-rw-r--r-- 3 thiago thiago    0 Sep 22 18:51 soft_link_file
```

I observed that when I changed `myfile` inserting content on it, the `module` made exactly what I was expecting:

```
BEGIN 'filecheck_files_dirs_example.file_existence' 9999721
SET 'file_/tmp/go_d_tests/myfile_exists' = 1
SET 'file_/tmp/go_d_tests/soft_link_file_exists' = 1
SET 'file_/tmp/go_d_tests/hard_link_exists' = 1
END

BEGIN 'filecheck_files_dirs_example.file_mtime_ago' 9999721
SET 'file_/tmp/go_d_tests/myfile_mtime_ago' = 8
SET 'file_/tmp/go_d_tests/soft_link_file_mtime_ago' = 8
SET 'file_/tmp/go_d_tests/hard_link_mtime_ago' = 8
END

BEGIN 'filecheck_files_dirs_example.file_size' 9999721
SET 'file_/tmp/go_d_tests/myfile_size_bytes' = 459202
SET 'file_/tmp/go_d_tests/soft_link_file_size_bytes' = 459202
SET 'file_/tmp/go_d_tests/hard_link_size_bytes' = 459202
END

BEGIN 'filecheck_files_dirs_example.dir_existence' 9999721
SET 'dir_/tmp/go_d_tests/mydir_exists' = 1
SET 'dir_/tmp/go_d_tests/soft_link_exists' = 1
END

BEGIN 'filecheck_files_dirs_example.dir_mtime_ago' 9999721
SET 'dir_/tmp/go_d_tests/mydir_mtime_ago' = 690
SET 'dir_/tmp/go_d_tests/soft_link_mtime_ago' = 690
END

BEGIN 'filecheck_files_dirs_example.dir_num_of_files' 9999721
SET 'dir_/tmp/go_d_tests/mydir_num_of_files' = 0
SET 'dir_/tmp/go_d_tests/soft_link_num_of_files' = 0
END

BEGIN 'netdata.execution_time_of_filecheck_files_dirs_example' 9999721
SET 'time' = 0
END
```

But, when I removed the file `my_file` I saw this content:

```
BEGIN 'filecheck_files_dirs_example.file_size' 9999990
SET 'file_/tmp/go_d_tests/myfile_size_bytes' = 
SET 'file_/tmp/go_d_tests/soft_link_file_size_bytes' = 1
SET 'file_/tmp/go_d_tests/hard_link_size_bytes' = 1
END
```

I think I am missing something, but I was expecting the dimension would receive `mx[fileDimID(filepath, "exists")] = 0` that would avoid it to be printed, or we would send an additional information for Netdata saying that the dimension is not more present.

ilyam8:
What is wrong here? No file - no size.

```
BEGIN 'filecheck_files_dirs_example.file_size' 9999990
SET 'file_/tmp/go_d_tests/myfile_size_bytes' = 
SET 'file_/tmp/go_d_tests/soft_link_file_size_bytes' = 1
SET 'file_/tmp/go_d_tests/hard_link_size_bytes' = 1
END
```

thiagoftsm (approved):
The PR is working as expected. I tested it using the reported structure and doing the following tests:

- First test: I set directories, files and links, and the configuration file had the correct specification for each one of these files.
- Second test: I changed the configuration file moving the  files to directory section and vice versa. The collector identified the "mistake" and it did not move forward.

LGTM.

ilyam8:
> the configuration file moving the files to directory section and vice versa

All such scenarios (add/remove files/dirs during runtime) should be in tests, testing manually is not good. We will add them later :trollface:

thiagoftsm:
> > the configuration file moving the files to directory section and vice versa
> 
> All such scenarios (add/remove files/dirs during runtime) should be in tests, testing manually is not good. We will add them later :trollface:

I agree, but I have to understand the whole collector, so the tests were necessary! :smile:

vlvkobal (changes requested):

vlvkobal (approved):

Fran17mc:
Hello, I'm trying use this plugin, but I think it isn't working properly.

My config, using:
```
cd /etc/netdata # Replace this path with your Netdata config directory
sudo ./edit-config go.d/filecheck.conf

```
Is:
```
# ------MODULE-CONFIGURATION--------
#  GLOBAL 
 update_every: 10
 autodetection_retry: 0
 priority: 70000
#
#
#  JOBS 
jobs:
 - name: fileMonitor
   files:
     include:
       - '/root/trees/soccer.pkl'
       - '/root/trees/basket.pkl'

```

Testing it with this code:
```
cd /usr/libexec/netdata/plugins.d/
sudo -u netdata -s

./go.d.plugin -d -m filecheck
```

The output is:

```
[ INFO  ] main[main] agent.go:106 instance is started
[ INFO  ] main[main] setup.go:39 loading config file
[ INFO  ] main[main] setup.go:47 looking for 'go.d.conf' in [/etc/netdata /usr/lib/netdata/conf.d]
[ INFO  ] main[main] setup.go:54 found '/etc/netdata/go.d.conf
[ INFO  ] main[main] setup.go:61 config successfully loaded
[ INFO  ] main[main] agent.go:110 using config: enabled 'true', default_run 'true', max_procs '0'
[ INFO  ] main[main] setup.go:66 loading modules
[ INFO  ] main[main] setup.go:85 enabled/registered modules: 1/49
[ INFO  ] main[main] setup.go:90 building discovery config
[ INFO  ] main[main] setup.go:116 looking for 'filecheck.conf' in [/etc/netdata/go.d /usr/lib/netdata/conf.d/go.d]
[ INFO  ] main[main] setup.go:123 found '/etc/netdata/go.d/filecheck.conf
[ INFO  ] main[main] setup.go:128 dummy/read/watch paths: 0/1/0
[ INFO  ] discovery[manager] manager.go:90 registered discoverers: [file discovery: [file reader]]
[ INFO  ] discovery[manager] manager.go:95 instance is started
[ INFO  ] run[manager] run.go:30 instance is started
[ INFO  ] build[manager] build.go:106 instance is started
[ INFO  ] discovery[file manager] discovery.go:71 instance is started
[ INFO  ] discovery[file reader] read.go:39 instance is started
[ INFO  ] discovery[file reader] read.go:40 instance is stopped
[ DEBUG ] run[manager] run.go:41 tick 0
[ DEBUG ] run[manager] run.go:41 tick 1
[ DEBUG ] run[manager] run.go:41 tick 2
[ DEBUG ] run[manager] run.go:41 tick 3
[ DEBUG ] run[manager] run.go:41 tick 4
[ DEBUG ] run[manager] run.go:41 tick 5
[ DEBUG ] run[manager] run.go:41 tick 6
[ DEBUG ] run[manager] run.go:41 tick 7
....
^C[ INFO  ] main[main] agent.go:191 received interrupt signal (2). Terminating...
```

It continue without any more information, and I can't show anything in my cloud dashboard. 

What could be the problem?

Thanks!

OdysLam:
Hey @Fran17mc 

Welcome to our little community :) One of the Agent engineers will come on this shortly.  The interruption at the end is due to your action, right? It appears to be working normally.

ilyam8:
~ solved in https://github.com/netdata/go.d.plugin/issues/473

Issue #442, opened 2020-09-20
[mysql] missing userstats_cpu chart


kam1kaze:
Hi,

I've just migrated from python mysql plugin to go and found that it does not contain `userstats_cpu` chart. I've checked the go code and have not found anything about it. Looks like it was missed.

Python chart is here:
https://github.com/netdata/netdata/blob/master/collectors/python.d.plugin/mysql/mysql.chart.py#L237

ilyam8:
Indeed, i forgout about it because it was a separate chart - i added only per user charts

https://github.com/netdata/go.d.plugin/blob/b1e849a145abe8e620e49fb9b884ce70ef564d6e/modules/mysql/charts.go#L706-L738

kam1kaze:
@ilyam8 thanks for the answer. Per user charts work fine.

ilyam8:
Do you think we need to add it as aggregated chart (as in py) and make it a chart per user (like commands)?

kam1kaze:
I would like to see it as aggregated chart (as in py), bcz I use it to determinate which services (client) are responsible for excessive load.

PR #446, opened 2020-09-22
modules/mysql: collect user statistics `Cpu_time`

ilyam8:
Fixes: #442

> The cumulative CPU time elapsed while servicing this user's connections.

```
Cpu_time: 0.0021376
```

Python [`userstas_cpu` chart](https://github.com/netdata/netdata/blob/83c0121c53c25c58e4308e9723f5395af29ef9c0/collectors/python.d.plugin/mysql/mysql.chart.py#L667-L671) units is `percentage`.

I dont get how we calc percentage 🤔 

https://github.com/netdata/netdata/blob/83c0121c53c25c58e4308e9723f5395af29ef9c0/collectors/python.d.plugin/mysql/mysql.chart.py#L961-L962

---

ok i got it - we need `mul` 100, `div` update_every and `algorithm` incremental.


ilyam8:
@kam1kaze could you help here? i think python `userstats_cpu` chart is wrong

vlvkobal (approved):

thiagoftsm (approved):
LGTM!

kam1kaze:
@ilyam8 as you wrote we need to multiply by 100. Does python script not have it? I thought that 4th param is multiplicator.

ilyam8:
> Does python script not have it?

Yeah, it does it, you are right.

I see python:
- doesnt take into account that `Cpu_time` is `float` and we need to multiply it by some precision and add this precision as divisor (netdata's external plugins api allows only `int` values).
- doesnt take into accout `update_every` so it works ~correctly only when `update_every` is 1.

PR #450, opened 2020-09-26
agent: exit 0 on SIGPIPE

ilyam8:
Netdata daemon doesnt restart plugin if its exit code is not 0 (even if it has been sending data for a while).

I did some tests on k8s and somehow/sometimes `go.d.plugin` got SIGPIPE on some nodes after fresh install, netdata was running tho. I think we need to exit 0 and allow netdata to restart the plugin.


thiagoftsm (approved):
LGM!

PR #451, opened 2020-09-29
ISC DHCPD collector.

thiagoftsm:
The test is completely commented, because it is generating an error.


modules/isc_dhcpd/lease.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"errors"
+	"net"
+	"os"
+	"time"
+
+	"github.com/netdata/go.d.plugin/agent/module"
+	leases "github.com/npotts/go-dhcpd-leases"
+)
```
ilyam8, line 11: i dont think we need `github.com/npotts/go-dhcpd-leases`

Please make `parse` a separate file, it contains

```go
package isc_dhcpd

type Lease struct {
	IP           string
	Ends         string
	BindingState string
}

func parseDHCPDLeases(leases []Lease, filepath string) ([]Lease, error) {
	...
}

```

It should have **only this logic, no any additional**. And add `parse_test.go` file and put here tests for different inputs (files/cases:ipv4 leases, ipv6 leases, backup leases (has no ends), static leases - end is `never`, etc)

---

`parseDHCPDLeases` takes `[]Lease` and returns `[]Lease`, yes, this is needed to reuse same slice.

`Ends` is string because sometimes it is `never`.

ilyam8 (changes requested):

go.mod
```diff
@@ -17,7 +17,9 @@ require (
 	github.com/mattn/go-isatty v0.0.12
 	github.com/miekg/dns v1.1.29
 	github.com/mitchellh/go-homedir v1.1.0
+	github.com/netdata/go-orchestrator v0.0.0-20200917191103-724cb40df90f
```
ilyam8, line 20: we dont use `go-orchestrator` anymore, remove all imports

modules/isc_dhcpd/isc_dhcpd.go
```diff
@@ -0,0 +1,97 @@
...
+		Dim       map[string]Dimensions
+	}
+
+	Dimensions struct {
+		Values ip.IRange
+		Name  string
+	}
+)
+
+type DHCPD struct {
```
ilyam8, line 22: DHCP daemon, from what i see people name it `DHCPd`, not `DHCPD`

modules/isc_dhcpd/isc_dhcpd.go
```diff
@@ -0,0 +1,97 @@
...
+	Config `yaml:",inline"`
+
+	collectedLeases bool
+	charts          *module.Charts
+}
+
+func init() {
+	module.Register("isc_dhcpd", module.Creator{
+		Defaults: module.Defaults{
+			UpdateEvery: 10,
```
ilyam8, line 33: we should use update_every 1, but to not parse file every second we need to the file check modification time, if it is changed - we parse it.

thiagoftsm: I already did a partial change in this direction and I am checking the file before to parse it. After to rewrite the parser, everything will be completed.

modules/isc_dhcpd/lease.go
thiagoftsm: I changed the parser, now we have our parser doing what we need.

modules/isc_dhcpd/isc_dhcpd_test.go
```diff
@@ -0,0 +1,280 @@
+package isc_dhcpd
+
+import (
+	"testing"
+	"net"
+
+	"github.com/netdata/go.d.plugin/pkg/ip"
```
ilyam8, line 7: rebase and use `iprange` pkg

modules/isc_dhcpd/lease.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func ParseDHCPd(r io.Reader) ([]LeaseFile) {
+	var list []LeaseFile
+	set := make(map[string]int)
+
+	scanner := bufio.NewScanner(r)
+	l := LeaseFile{}
+	for scanner.Scan() {
+		cmp := bytes.TrimSpace(scanner.Bytes())
```
ilyam8, line 28: call it line, not cmp

modules/isc_dhcpd/lease.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+type LeaseFile struct {
+	IP    net.IP
+	Ends  time.Time
+	EndString string
+	State string
+}
+
+func ParseDHCPd(r io.Reader) ([]LeaseFile) {
```
ilyam8, line 21: `LeaseFile` => `lease` or `leaseEntry`

all fields should be private, why public?

`State` => `bindingState` because it is binding state.

---

shouldnt be public and should be 

`parseDHCPdLeases(leases []lease, r io.Reader) ([]lease)`

ilyam8 (changes requested):

modules/isc_dhcpd/isc_dhcpd_test.go
thiagoftsm: I finished the rebase.

modules/isc_dhcpd/lease.go
thiagoftsm: Fixed with the last commit.

modules/isc_dhcpd/lease.go
thiagoftsm: > all fields should be private, why public?

I forgot the detail that `go` convert to public when variable is Upper case .

thiagoftsm: I changed everything according we talked.

ilyam8:
@thiagoftsm i think we don't need `ends`, because we want to reflect `dhcpd` real state which is `dhcpd.leases` file, if `binding state` is `active` then it is `active` leases (i understand that if `dhcpd` daemon is stopped then the client lease database is not updated - we show active leases which are not active in fact, but it is ok).

please finish the PR:
- update config
- update tests
- other fixes if you think there is something i missed or did in a wrong way, lets discuss

modules/isc_dhcpd/isc_dhcpd_test.go
```diff
@@ -157,3 +185,73 @@ func ensureCollectedHasAllChartsDimsVarsIDs(t *testing.T, dhcpd *DHCPd, collecte
 		}
 	}
 }
+
+func leaseDoesNotExist() *DHCPd {
```
ilyam8, line 189: call those funtions `prepareDHCPd...` where `...` is the test case.

and it is not `lease` it is `LeaseDB`

> dhcpd.leases - DHCP client lease database

ilyam8 (changes requested):

ilyam8:
Lease db is `dhcpd.leases` lets call files like `dhcpd.leases_SUFFIX`:
- `dhcpd.leases_ipv4`
- `dhcpd.leases_ipv4_backup`
- `dhcpd.leases_ipv4_empty` (not clean)
- etc.

What do you think?

modules/isc_dhcpd/isc_dhcpd_test.go
```diff
@@ -43,42 +78,33 @@ func TestDHCPd_Init(t *testing.T) {
...
 	}
 
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
 			dhcpd := test.prepare()
-			require.True(t, dhcpd.Init())
+			dhcpd.Init()
 
 			if test.wantFail {
 				assert.False(t, dhcpd.Init())
```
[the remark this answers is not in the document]
thiagoftsm: Considering the tests you propose, the system will fail when it tries to open a file that does not exist, this was the motive I moved the `Init` outside the tests.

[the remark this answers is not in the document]
ilyam8: it doesn't fail in `Init` that is why it was `require.True(t, dhcpd.Init())`, we want to ensure Init doesnt fail before the Check test case starts.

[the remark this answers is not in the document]
ilyam8: `test.wantFail` means we want `Check` fail or not (false/true).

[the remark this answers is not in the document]
ilyam8: or man, are we testing `Init` here? 🤣 then nvm lol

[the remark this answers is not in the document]
ilyam8: i was under imression it is Check tests!

thiagoftsm:
> What do you think?

I agree with you, the last commit brings the files renamed.

I also renamed the function as you required.

modules/isc_dhcpd/isc_dhcpd_test.go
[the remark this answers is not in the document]
ilyam8: well, i rechecked, yes, these are `Check` tests! so we need the job inited before the test case - require.

ilyam8:
@thiagoftsm 
 - fix Check tests
 - fix `isc_dhcpd.conf`
 - run `go fmt ./...`

modules/isc_dhcpd/isc_dhcpd_test.go
```diff
@@ -0,0 +1,257 @@
...
+				Networks: "192.168.0.0/24",
+			},
+		},
+	}
+	
+	return dhdcpd
+}
+
+func prepareDHCPdIPV6Lease() *DHCPd {
+	dhdcpd := New()
```
ilyam8, line 245: dhdcpd => dhcpd (in all prepare* functions)

and add a new line to the end of the file

modules/isc_dhcpd/isc_dhcpd_test.go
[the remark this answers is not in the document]
thiagoftsm: All right, the problem was fixed :laughing:

modules/isc_dhcpd/isc_dhcpd_test.go
```diff
@@ -104,7 +104,7 @@ func TestDHCPd_Check(t *testing.T) {
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
 			dhcpd := test.prepare()
-			dhcpd.Init()
+			require.True(t, dhcpd.Init())
 
 			if test.wantFail {
 				assert.False(t, dhcpd.Init())
```
ilyam8, line 110: `assert.False(t, dhcpd.Init())` => `assert.False(t, dhcpd.Check())`, same for `else` branch

thiagoftsm:
> @thiagoftsm
> 
>     * fix Check tests
> 
>     * fix `isc_dhcpd.conf`
> 
>     * run `go fmt ./...`

Everything was made with the last commit.

ilyam8:
https://github.com/netdata/go.d.plugin/blob/isc_dhcp/config/go.d/isc_dhcpd.conf

it doesn't reflect https://github.com/netdata/go.d.plugin/blob/345c32933e4b9b9ce0040ccd57cbfecbcc8684c5/modules/isc_dhcpd/isc_dhcpd.go#L18-L27

thiagoftsm:
> https://github.com/netdata/go.d.plugin/blob/isc_dhcp/config/go.d/isc_dhcpd.conf
> 
> it doesn't reflect
> 
> https://github.com/netdata/go.d.plugin/blob/345c32933e4b9b9ce0040ccd57cbfecbcc8684c5/modules/isc_dhcpd/isc_dhcpd.go#L18-L27

I updated the `isc_dhcpd.conf` now according the documentation.

config/go.d/isc_dhcpd.conf
```diff
@@ -81,8 +84,8 @@
...
 #jobs:
 # - name: lease_example
-#   files:
-#     include:
-#       - '/path/to/lease_file1'
-#       - '/path/to/lease_file2'
+#   leases_path: '/path/to/ipv4_leases_file'
+#   pools:
+#       - name : example
+#           networks: "192.168.0.0/24 192.168.1.1-192.168.1.63 2001:0DB8::/64"
```
ilyam8: just copy both examples from the README.md

config/go.d/isc_dhcpd.conf
```diff
@@ -59,17 +59,20 @@
 #  - lease
```
ilyam8, line 59: what is that `lease` option?

config/go.d/isc_dhcpd.conf
```diff
@@ -82,10 +82,9 @@
...
-#   pools:
-#       - name : example
-#           networks: "192.168.0.0/24 192.168.1.1-192.168.1.63 2001:0DB8::/64"
-#
+jobs:
+ - name: lease_example
+   leases_path: '/path/to/ipv4_leases_file'
+   pools:
+       - name : example
+           networks: "192.168.0.0/24 192.168.1.1-192.168.1.63 2001:0DB8::/64"
```
ilyam8, line 90: should be commented out because it is an example, there is no reason to have this job with `leases_path: '/path/to/ipv4_leases_file'` and random pools.

ilyam8:
Documentation/examples: try to use ip addresses from the documentation address blocks:
 - v4: https://tools.ietf.org/html/rfc5737
 - v6: https://www.apnic.net/get-ip/faqs/ipv6-documentation-prefix/

config/go.d/isc_dhcpd.conf
thiagoftsm: I removed the comments, because `k8s_kuberlet.conf` has the final part without comments.

ilyam8: it has this predefined job because it makes sense to have it, there are bunch of modules that have predefined jobs.

The thing is sense - does it make sense to have it or net. It makes sense when we expect the job to be running with the default parameters. 

Example:

```yaml
# apache
jobs:
 - name: local
   url: "http://127.0.0.1/server_status?auto"
```

it is not commented out because we want to try `http://127.0.0.1/server_status?auto`, if there is apache is running, we get metrics from it.

---

In `isc_dhcpd` example there is no sense at all, that is why we commented out its jobs, we mainly add them as examples.

thiagoftsm: I got it.

ilyam8:
I updated test cases and readme/config.

ilyam8 (approved):

vlvkobal (approved):

PR #452, opened 2020-09-29
pkg/iprange: refactor and Pool support

ilyam8:


pkg/iprange/parse.go
```diff
@@ -0,0 +1,144 @@
...
+func isV6RangeValid(start, end net.IP) bool {
+	return isV6IP(start) && isV6IP(end) && bytes.Compare(end, start) >= 0
+}
+
+func isV4IP(ip net.IP) bool {
+	return ip.To4() != nil
+}
+
+func isV6IP(ip net.IP) bool {
+	return !isV4IP(ip) && ip.To16() != nil
```
thiagoftsm, line 143: A minor thing here, you tested the return from `isV4IP`  using a different pattern in the whole code until now. This comparison is not wrong, but it is not following the pattern.

pkg/iprange/parse_test.go
```diff
@@ -0,0 +1,258 @@
...
+				prepareV6Range("2001:db8::1", "2001:db8::2"),
+				prepareV4Range("192.0.2.1", "192.0.2.254"),
+			},
+		},
+		"single invalid syntax": {
+			input:   "192.0.2.0-192.0.2.",
+			wantErr: true,
+		},
+		"multiple invalid syntax": {
+			input:   "2001:db8::0 192.0.2.0-192.0.2.10 2001:db8::0/999 192.0.2.0/255.255.255.0",
```
thiagoftsm, line 37: please, add in the `invalid range`, a range with reverse order or change `192.0.2.0-192.0.2.10` to `192.0.2.10-192.0.2.0`.

thiagoftsm: You also could add an invalid mask, for example, instead `192.0.2.0/255.255.255.0`, you could use `255.255.255.0/192.0.2.0`

pkg/iprange/parse.go
ilyam8: what pattern?

pkg/iprange/parse_test.go
thiagoftsm: I saw now that the next function has the tests I am suggesting here.

ilyam8: both of these cases are tested in `ParseRange`

start > end
https://github.com/ilyam8/go.d.plugin/blob/11eacd11ecb077277d33de799364d2084afd0847/pkg/iprange/parse_test.go#L79-L81

invalid/not canonical form mak

https://github.com/ilyam8/go.d.plugin/blob/11eacd11ecb077277d33de799364d2084afd0847/pkg/iprange/parse_test.go#L151-L158

---

i dont need to test these cases in the `ParseRanges`, i test only its gist - it parses single and multiple ranges (exact ranges types doesnt matter)

pkg/iprange/parse_test.go
```diff
@@ -0,0 +1,258 @@
...
+			wantRange: prepareV4Range("192.0.2.1", "192.0.2.2"),
+		},
+		"v4 Mask: /31": {
+			input:     "192.0.2.0/255.255.255.254",
+			wantRange: prepareV4Range("192.0.2.0", "192.0.2.1"),
+		},
+		"v4 Mask: /32": {
+			input:     "192.0.2.0/255.255.255.255",
+			wantRange: prepareV4Range("192.0.2.0", "192.0.2.0"),
+		},
```
thiagoftsm, line 146: Please, add here an inverted IP/MASK pair like `255.255.255.254/192.0.2.0`

ilyam8: it is not inverted mask, it is not canonical form mask (ones followed by zeros)

ilyam8: this is tested

pkg/iprange/range.go
```diff
@@ -0,0 +1,85 @@
...
+	size := big.NewInt(0)
+	size.Add(size, big.NewInt(0).SetBytes(r.end))
+	size.Sub(size, big.NewInt(0).SetBytes(r.start))
+	size.Add(size, big.NewInt(1))
+	return size
+}
+
+func v4ToInt(ip net.IP) int64 {
+	ip = ip.To4()
+	return int64(ip[0])<<24 | int64(ip[1])<<16 | int64(ip[2])<<8 | int64(ip[3])
```
thiagoftsm, line 84: If Netdata runs on `big endian` machine, this inversion is not necessary. I think we should test here the endianness before to return, instead to assume we are only running on little Endian.

ilyam8: do you have such machine to test this function?

pkg/iprange/parse.go
thiagoftsm: You wrote for the majority of the code `function(value) != nil` instead `!function(value)`.

ilyam8: it is not pattern, i could write it as `ip.To4() == nil && ip.To16() != nil` but since i have `isV4IP` i used it

std lib source code
```go
// To4 converts the IPv4 address ip to a 4-byte representation.
// If ip is not an IPv4 address, To4 returns nil.
func (ip IP) To4() IP {
	...
}

// To16 converts the IP address ip to a 16-byte representation.
// If ip is not an IP address (it is the wrong length), To16 returns nil.
func (ip IP) To16() IP {
	...
}
```

So `To16()` returns **not nil** for both v4 and v6, `To4()` returns **not nil** only for v4.

pkg/iprange/range.go
thiagoftsm: I do not have `bigendian` computers, but I saw codes like this [here](https://github.com/lattera/glibc/blob/master/string/endian.h) and [here](https://github.com/lattera/glibc/blob/master/bits/uintn-identity.h).

Considering that Network IPs are `bigendian`, I assumed you are converting them to `little endian`

thiagoftsm: After we talk on Slack, I could solve the confusion with bits and text.

thiagoftsm (approved):
LGTM!

vlvkobal (approved):

PR #461, opened 2020-10-17
[WIP] Add wireguard support

lets00:
# Summary

This PR adds wireguard collector. It collects data about peers (connected or no) in a interface Wireguard VPN.


modules/wireguard/wireguard.go
```diff
@@ -0,0 +1,128 @@
...
+	w.metrics["sent_total"] = 0
+
+	device, _ := w.connection.Device(w.Interface)
+
+	for id, peer := range device.Peers {
+		receivedKey := fmt.Sprintf("received_%v", id)
+		sentKey := fmt.Sprintf("sent_%v", id)
+
+		w.metrics[receivedKey] = (peer.ReceiveBytes - w.lastValues[receivedKey]) * 8 / w.UpdateEvery
+		w.metrics[sentKey] = (peer.TransmitBytes - w.lastValues[sentKey]) * 8 / w.UpdateEvery
```
ilyam8, line 119: make dimensions algoriths `Incremental` and you dont need to save last values, etc. Just provide the values as is.

ilyam8:
Hi @lets00 long time no see 😄

lets00:
Yes @ilyam8 ! I'm coming back collaborate with this amazing project.

modules/wireguard/wireguard.go
lets00: Ok. I'll fix this.

lets00: But, I want the difference between the actual value and last (because peer.ReceiveBytes always return total bytes value). How can I do this using `Algo: module.Incremental` ?

ilyam8: just returns value as is, just make it `module.Incremental`, yes.

See https://learn.netdata.cloud/docs/agent/collectors/plugins.d#dimension (algorithm->incremental)

ilyam8:
@lets00 i think [systemdunits](https://github.com/netdata/go.d.plugin/tree/master/modules/systemdunits)  collector will help you to write tests, looks like you need to use same logic - use `wgClient` and `wgConnection` interfaces.

modules/wireguard/wireguard.go
lets00: Hmm, too simple! I'll fix this.

lets00:
Hi @ilyam8 , What do I need to do to see the charts in netdata UI? I copied the binary to plugins.d(````cp bin/goplugin /usr/libexec/netdata/plugins.d/go.d.plugin````)and add config file in /etc/netdata (```cp -r config/* /etc/netdata```) . I ran the collector test command (```./go.d.plugin -d -m wireguard```), It showed that my collector is up and running, but in UI the charts doesn't appear. Did I forget any important step?

ilyam8:
Hi. See `error.log`

> grep wireguard error.log

Could be systemd unit permissions issue

ilyam8:
@lets00 click on `Ready for review` when it is ready.

---

https://github.com/netdata/go.d.plugin/pull/461#issuecomment-713514322 so what was the problem? Or it is still the problem?

lets00:
Ok @ilyam8. When it will be ready I click in ready for review.

The problem was permission. When I defined the correct permission the collector works (it starts to collect wireguard's data).

go.mod
```diff
@@ -27,7 +28,7 @@ require (
 	github.com/tomasen/fcgi_client v0.0.0-20180423082037-2bb3d819fd19
 	github.com/valyala/fastjson v1.5.4
 	github.com/vmware/govmomi v0.22.2
-	golang.org/x/net v0.0.0-20191027093000-83d349e8ac1a // indirect; needed for freebsd/arm64
```
lets00, line 30: @ilyam8 . I don't know why this package was removed when I was working on wireguard module. Should I include it?

ilyam8: as you see it is needed for `freebsd/arm64`

try w/o it

```cmd
0 go.d.plugin (master)$ pwd
/Users/ilyam/Projects/golang/go.d.plugin

0 go.d.plugin (master)$ hack/go-build.sh freebsd/arm64
Building binaries for version: v0.24.0
Building freebsd/arm64
```

liepumartins (approved):
Successfully testing on one instance.

Probably should also edit `apps_groups.conf` and add `wireguard` to `vpn` (but that goes with main netdata)

ilyam8:
> Probably should also edit apps_groups.conf and add wireguard to vpn (but that goes with main netdata)

@liepumartins actually a good idea 👍 are you willing to make a PR?

modules/wireguard/README.md
```diff
@@ -0,0 +1,53 @@
...
+This module will monitor one or more vpn network interface and wireguard peers, depending on your configuration.
+
+## Requirements
+
+- `Wireguard` enabled
+- Set network capability to golang collector (default file is: `/usr/libexec/netdata/plugins.d/go.d.plugin`)
+
+E.g:
+```bash
+$ sudo setcap CAP_NET_ADMIN+epi /usr/libexec/netdata/plugins.d/go.d.plugin
```
liepumartins, line 14: FYI On a systemd system I found that I had to alter systemd unit file and add `CapabilityBoundingSet=CAP_NET_ADMIN` there.
Setting capability alone was not enough and threw errors like `Operation not permitted` in `error.log`

ilyam8: You are right, we need to update netdata systemd unit CapabilityBoundingSet.

>  to alter systemd unit file

No need to alter the file.

[(source)](https://www.freedesktop.org/software/systemd/man/systemd.unit.html)

> Along with a unit file foo.service, a "drop-in" directory foo.service.d/ may exist. All files with the suffix ".conf" from this directory will be parsed after the unit file itself is parsed. This is useful to alter or add configuration settings for a unit, without having to modify unit files.


So something like this will do

```cmd
mkdir /etc/systemd/system/netdata.service.d
echo -e '[Service]\nCapabilityBoundingSet=CAP_NET_ADMIN' | tee /etc/systemd/system/netdata.service.d/unset-capability-bounding-set.conf
systemctl daemon-reload
systemctl restart netdata.service
```

PR #462, opened 2020-10-20
convert histogram options to microseconds

hamedbrd:
##### Summary

Both `request processing time` and `upstream response time` values are in microseconds, so we need to convert histogram options to microseconds (ms=>us) and then pass it to NewHistogram in order to show the histogram options in charts 

##### Additional Information
Fixes [#10075](https://github.com/netdata/netdata/issues/10075)


modules/weblog/metrics.go
```diff
@@ -130,3 +132,12 @@ func newReqCustomField(fields []customField) map[string]metrics.CounterVec {
 	}
 	return cf
 }
+
+// convert histogram options to microseconds (ms => us)
+func convHistOptionsToMicroseconds(histogram []float64) []float64 {
+	var buckets []float64
+	for _, value := range histogram {
+		buckets = append(buckets, float64(time.Duration(value)*time.Microsecond))
```
ilyam8, line 140: I think we can just `* 1000`.

```suggestion
		buckets = append(buckets, value * 1000)
```

hamedbrd: it was like this, I thought maybe would be better to not hard code 1000.

ilyam8:
> modules/weblog/metrics.go:4:2: imported and not used: "time"

ilyam8:
And there is another chart we need to fix units

https://github.com/netdata/go.d.plugin/blob/1305e1afb4b57776ce29daa2fedfa631c16dde1d/modules/weblog/charts.go#L230-L237

ilyam8:
I see default bucket is in seconds

https://github.com/netdata/go.d.plugin/blob/1305e1afb4b57776ce29daa2fedfa631c16dde1d/pkg/metrics/histogram.go#L38-L42

I think we need to change `histogram` option and make it `in seconds`, not `in milliseconds`

https://github.com/netdata/go.d.plugin/blob/1305e1afb4b57776ce29daa2fedfa631c16dde1d/config/go.d/web_log.conf#L98-L101

Then we need multiplier `1e6`

hamedbrd:
@ilyam8  do you know why tests failed?
https://app.circleci.com/pipelines/github/netdata/go.d.plugin/2245/workflows/e04c9cc6-aeb1-4390-8e54-477acf175339/jobs/7224

ilyam8:
Yes, i know, lets fix https://github.com/netdata/go.d.plugin/pull/462#issuecomment-712820708 and https://github.com/netdata/go.d.plugin/pull/462#issuecomment-712828995, then we fix the tests

hamedbrd:
@ilyam8 shall I change the `histogram` options in `weblog.conf ` to like what we in `DefBuckets` ?

ilyam8:
@hamedbrd yeah

config/go.d/web_log.conf
```diff
@@ -98,7 +98,7 @@
 #  - histogram
 #    Cumulative histogram of response time in milli seconds.
```
ilyam8, line 99:
```suggestion
#    Cumulative histogram of response time in seconds.
```

hamedbrd:
Screenshot 2020-10-20 at 16 26 40

ilyam8:
@hamedbrd you need to change values in `expected` map (TestWebLog_Collect), see `-`(Expected)/`+`(Actual)

https://app.circleci.com/pipelines/github/netdata/go.d.plugin/2248/workflows/79f0d7cf-042a-4fc7-980c-bcd344531e4d/jobs/7238

ilyam8:
And i was wrong, both those charts units should have `/s` suffix since we use `Incremental` dim algorithm 😅

ilyam8:
@hamedbrd 

https://github.com/netdata/go.d.plugin/pull/462#issuecomment-712898899 looks good

thx!

thiagoftsm (approved):
Thank you very much @hamedbrd !

LGTM!

PR #467, opened 2020-10-23
Add Custom time fields to WebLog Module

hamedbrd:
##### Summary

Add `custom_time_field` to `web_log` module to monitor custom time fields that are not supported by default. in the config file of web_log we need to add something like this to monitor those fields.
```
jobs:
  - name: apache
    ...
    custom_time_fields:
     - name:  '^FB'  # implying we use custom log format and %^FB is in there
       histogram: [.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10]  # optional
```

##### Additional Information
Fixes [#9611](https://github.com/netdata/netdata/issues/9611)

Screenshot 2020-10-23 at 16 57 51


modules/weblog/charts.go
```diff
@@ -360,6 +362,31 @@ var (
 	}
 )
 
+// custom time field
+var (
+	reqByCustomTimeField = Chart{
+		ID:       "custom_time_field_%s_requests",
+		Title:    "Custom Time Field %s Requests",
+		Units:    "requests/s",
```
ilyam8, line 370: I think it is not _Requests_, let's call it `custom_time_field_%s_summary`

`units` => `milliseconds`.
`context` => `custom_time_field_summary`

modules/weblog/charts.go
```diff
@@ -360,6 +362,31 @@ var (
...
+			{ID: "custom_time_field_%s_time_avg", Name: "avg", Div: 1000},
+		},
+	}
+	reqByCustomTimeFieldHist = Chart{
+		ID:       "custom_time_field_hist_%s_requests",
+		Title:    "Custom Time Field Histogram %s",
+		Units:    "requests/s",
+		Fam:      "custom time field hist",
+		Ctx:      "web_log.custom_time_hist_field_%s",
+		Priority: prioReqCustomTimeFieldHist,
```
ilyam8, line 386: same, `custom_time_field_hist_%s_requests` => `custom_time_field_%s_histogram`

`units` => `observations`
`context` => `custom_time_field_histogram`

modules/weblog/collect.go
```diff
@@ -312,18 +312,32 @@ func (w *WebLog) collectCustomFields() {
...
+			v, ok := w.mx.ReqCustomTimeField[cv.name]
 			if !ok {
 				break
 			}
-			c, _ := v.GetP(pattern.name)
-			c.Inc()
-			break
+			ctf, err := strconv.ParseFloat(cv.value, 64)
+			if err != nil || !isTimeValid(ctf) {
+				break
```
ilyam8, line 337: and here

modules/weblog/collect.go
```diff
@@ -312,18 +312,32 @@ func (w *WebLog) collectCustomFields() {
...
+				}
+				c, _ := v.GetP(pattern.name)
+				c.Inc()
+				break
 			}
-			v, ok := w.mx.ReqCustomField[cv.name]
+		} else if _, ok := w.customTimeFields[cv.name]; ok {
+			v, ok := w.mx.ReqCustomTimeField[cv.name]
 			if !ok {
 				break
```
ilyam8, line 333: looks like we need `continue` here

modules/weblog/metrics.go
```diff
@@ -131,6 +141,21 @@ func newReqCustomField(fields []customField) map[string]metrics.CounterVec {
...
+func newReqCustomTimeField(fields []customTimeField) map[string]*customTimeFieldMetrics {
+	cf := make(map[string]*customTimeFieldMetrics)
+	for _, f := range fields {
+		var buckets []float64
+		for _, value := range f.Histogram {
+			buckets = append(buckets, value*1e6)
+		}
+		cf[f.Name] = &customTimeFieldMetrics{
+			Time:     newWebLogSummary(),
+			TimeHist: metrics.NewHistogram(buckets),
```
ilyam8, line 153: Do we really need default bucket if histogram is not defined? I see you add histogram chart if len(f.Histogram) > 0, seems the answer is no

modules/weblog/collect.go
```diff
@@ -312,18 +312,32 @@ func (w *WebLog) collectCustomFields() {
...
 			}
-			c, _ := v.GetP(pattern.name)
-			c.Inc()
-			break
+			ctf, err := strconv.ParseFloat(cv.value, 64)
+			if err != nil || !isTimeValid(ctf) {
+				break
+			}
+			v.Time.Observe(ctf)
+			v.TimeHist.Observe(ctf)
```
ilyam8, line 340: Since histogram is optional `v.TimeHist` could be `nil` - atm it is not the case because you always create `v.TimeHist` (even if custom_time_field.field.histogram not set)

ilyam8:
@hamedbrd good job 👍

modules/weblog/collect.go
```diff
@@ -312,18 +312,32 @@ func (w *WebLog) collectCustomFields() {
...
 			if !ok {
 				break
 			}
-			c, _ := v.GetP(pattern.name)
-			c.Inc()
-			break
+			ctf, err := strconv.ParseFloat(cv.value, 64)
+			if err != nil || !isTimeValid(ctf) {
+				break
+			}
```
ilyam8, line 338: oh, we need time multiplier

https://github.com/netdata/go.d.plugin/blob/a1c1fa72aba66e0f9dbf7ed9de239b849b47b474/modules/weblog/logline.go#L315-L319

modules/weblog/collect.go
```diff
@@ -312,18 +312,35 @@ func (w *WebLog) collectCustomFields() {
...
 			}
-			v, ok := w.mx.ReqCustomField[cv.name]
+		} else if _, ok := w.customTimeFields[cv.name]; ok {
+			v, ok := w.mx.ReqCustomTimeField[cv.name]
 			if !ok {
+				continue
+			}
+			ctf, err := strconv.ParseFloat(cv.value, 64)
+			if err != nil || !isTimeValid(ctf) {
 				break
```
ilyam8, line 337:
```suggestion
				continue
```

modules/weblog/charts.go
```diff
@@ -537,6 +564,73 @@ func newCustomFieldChart(f customField) (*Chart, error) {
...
+		}
+	}
+	return charts, nil
+}
+
+func newCustomTimeFieldChart(f customTimeField) (*Chart, error) {
+	chart := reqByCustomTimeField.Copy()
+	chart.ID = fmt.Sprintf(chart.ID, f.Name)
+	chart.Title = fmt.Sprintf(chart.Title, f.Name)
+	chart.Ctx = fmt.Sprintf(chart.Ctx, f.Name)
```
ilyam8, line 585: > chart.Ctx = fmt.Sprintf(chart.Ctx, f.Name)

not needed

modules/weblog/charts.go
```diff
@@ -537,6 +564,73 @@ func newCustomFieldChart(f customField) (*Chart, error) {
...
+		d.ID = fmt.Sprintf(d.ID, f.Name)
+	}
+	return chart, nil
+}
+
+func newCustomTimeFieldHistChart(f customTimeField) (*Chart, error) {
+	chart := reqByCustomTimeFieldHist.Copy()
+	chart.ID = fmt.Sprintf(chart.ID, f.Name)
+	chart.Title = fmt.Sprintf(chart.Title, f.Name)
+	chart.Ctx = fmt.Sprintf(chart.Ctx, f.Name)
```
ilyam8, line 596: > chart.Ctx = fmt.Sprintf(chart.Ctx, f.Name)

not needed

ilyam8 (changes requested):

ilyam8:
@hamedbrd i think we don't need 2 families, it should be one `custom time field` and both summary and histogram charts in there.

i think this will do:

have both summary and histogram templates with same default priority and ...

example code

```go
func newCustomTimeFieldCharts(fields []customTimeField) (Charts, error) {
	charts := Charts{}
	for i, f := range fields {
		chart, err := newCustomTimeFieldChart(f)
		if err != nil {
			return nil, err
		}
        chart.Priority += i

		if err := charts.Add(chart); err != nil {
			return nil, err
		}

		chart, err := newCustomTimeFieldHistChart(f)
		if err != nil {
			return nil, err
		}
        chart.Priority += i + 1

		if err := charts.Add(chart); err != nil {
			return nil, err
		}
	}
	return charts, nil
}
```

ilyam8:
i mean like this (charts order)

```
custom_time_fields:
 - field1_summary
 - field1_histogram
 - field2_summary
 - field2_histogram
 ...
```

hamedbrd:
> i mean like this (charts order)
> 
> ```
> custom_time_fields:
>  - field1_summary
>  - field1_histogram
>  - field2_summary
>  - field2_histogram
>  ...
> ```

Do you mean something like this?
Screenshot 2020-10-28 at 17 07 16

modules/weblog/charts.go
```diff
@@ -360,6 +362,31 @@ var (
 	}
 )
 
+// custom time field
+var (
+	reqByCustomTimeField = Chart{
+		ID:       "custom_time_field_%s_summary",
+		Title:    "Custom Time Field %s Summary",
```
ilyam8, line 369:
```suggestion
		Title:    `Custom Time Field "%s" Summary`,
```

modules/weblog/charts.go
```diff
@@ -360,6 +362,31 @@ var (
...
+		Priority: prioReqCustomTimeField,
+		Dims: Dims{
+			{ID: "custom_time_field_%s_time_min", Name: "min", Div: 1000},
+			{ID: "custom_time_field_%s_time_max", Name: "max", Div: 1000},
+			{ID: "custom_time_field_%s_time_avg", Name: "avg", Div: 1000},
+		},
+	}
+	reqByCustomTimeFieldHist = Chart{
+		ID:       "custom_time_field_%s_histogram",
+		Title:    "Custom Time Field Histogram %s",
```
ilyam8, line 382:
```suggestion
		Title:    `Custom Time Field "%s" Histogram`,
```

ilyam8:
@hamedbrd the code looks ok 👍

We need to update the documentation https://github.com/netdata/go.d.plugin/tree/master/modules/weblog#custom-fields-feature

ilyam8:
@hamedbrd seems good, i think i will do some real tests to see how it looks on the dashboard and approve if everything is ok.

config/go.d/web_log.conf
```diff
@@ -95,6 +95,15 @@
 #          - name:  name2
 #            match: pattern
 #
+#  - custom_time_fields
+#    Count min/avg/max and also cumulative histogram for defined custom time fields like apache LogIOTrackTTFB . Used in custom log format.
```
vlvkobal, line 99:
```suggestion
#    Count min/avg/max and also cumulative histogram for defined custom time fields like apache LogIOTrackTTFB. Used in custom log format.
```

modules/weblog/README.md
```diff
@@ -261,6 +261,29 @@ we still can get some info from them.
...
 
+## Custom Time Fields Feature
+
+Weblog is also able to extract user defined time fields and could count min/avg/max + histogram against these fields.
+
+This feature needs:
+-   custom log format with user defined time fields
+-   histogram to show response time in seconds which is optional.
+
+There is an example in apache custom log forma with 1 custom time fields - `^FB` which shows delay in microseconds between when the request arrived and the first byte of the response headers are written. Like Custom Fields feature Weblog is unaware of these fields, but
```
vlvkobal, line 272: forma -> format

Please split this into several lines.

@joelhans please take a look.

modules/weblog/README.md
```diff
@@ -261,6 +261,31 @@ we still can get some info from them.
...
 
+## Custom Time Fields Feature
+
+Weblog is also able to extract user defined time fields and could count min/avg/max + histogram against these fields.
+
+This feature needs:
+-   custom log format with user defined time fields
+-   histogram to show response time in seconds, which is optional.
+
+As an example, apache has a custom log format that could be enabled by adding - `^FB` to the log format. This field shows a delay in microseconds between when the request arrived, and the first byte of the response headers are written.
```
ilyam8, line 272:
```suggestion
As an example, apache [`mod_logio`](https://httpd.apache.org/docs/2.4/mod/mod_logio.html) adds `^FB` logging directive. This value shows a delay in microseconds between when the request arrived, and the first byte of the response headers are written.
```

modules/weblog/init.go
```diff
@@ -69,11 +69,33 @@ func (w *WebLog) createCustomFields() error {
...
+	}
+
+	w.Debug("starting custom time fields creating")
+	w.customTimeFields = make(map[string][]float64)
+	for i, ctf := range w.CustomTimeFields {
+		if ctf.Name == "" {
+			return fmt.Errorf("create custom field: name not set (field %d)", i+1)
+		}
+		w.customTimeFields[ctf.Name] = ctf.Histogram
+		w.Debugf("created time field '%s', histogram '%T'", ctf.Name, ctf.Histogram)
```
ilyam8, line 85:
```suggestion
		w.Debugf("created time field '%s', histogram '%v'", ctf.Name, ctf.Histogram)
```

ilyam8: > [ DEBUG ] web_log[nginx] init.go:85 created time field 'time1', histogram '[]float64'
[ DEBUG ] web_log[nginx] init.go:85 created time field 'time2', histogram '[]float64

We are not interested in type but in actual value.

ilyam8 (approved):

ilyam8:
@hamedbrd thanks!

modules/weblog/README.md
vlvkobal: I meant that you shouldn't use very long lines. You should try to keep every line less than 120 characters.

thiagoftsm (approved):
Thank you @hamedbrd !

LGTM!

modules/weblog/README.md
```diff
@@ -154,14 +154,14 @@ In addition to that weblog understands [user defined fields](#custom-fields-feat
 Notes:
 
 -   Apache `%h` logs the IP address if [HostnameLookups](https://httpd.apache.org/docs/2.4/mod/core.html#hostnamelookups) is Off.
-    Weblog counts hostname as IPv4 address. We recommend either to disable HostnameLookups or use `%a` instead of `%h`. 
+    Weblog counts hostname as IPv4 address. We recommend either to disable HostnameLookups or use `%a` instead of `%h`.
```
joelhans, line 157: It's in a few other places in this doc, so we don't have to change them all now, but I think we should avoid calling it **Weblog**. I always read it first as "we blog." I think we should stick with these usages:

- The web log collector...
- Netdata's web log collector...
- The `web_log` collector...

```suggestion
    The web log collector counts hostnames as IPv4 addresses. We recommend either to disable HostnameLookups or use `%a` instead of `%h`.
```

modules/weblog/README.md
```diff
@@ -261,6 +261,31 @@ we still can get some info from them.
             match: '* *'
 ```
 
+## Custom Time Fields Feature
+
+Weblog is also able to extract user defined time fields and could count min/avg/max + histogram against these fields.
```
joelhans, line 266:
```suggestion
The web log collector is also able to extract user defined time fields and could count min/avg/max + histogram against these fields.
```

modules/weblog/README.md
```diff
@@ -261,6 +261,31 @@ we still can get some info from them.
...
+
+This feature needs:
+-   custom log format with user defined time fields
+-   histogram to show response time in seconds, which is optional.
+
+As an example, apache [`mod_logio`](https://httpd.apache.org/docs/2.4/mod/mod_logio.html) adds `^FB` logging directive. This value shows a delay in microseconds between when the request arrived, and the first byte of the response headers are written.
+
+Like Custom Fields feature Weblog is unaware of these fields, but
+we can still get some info from them.
```
joelhans, line 275:
```suggestion
As with the custom fields feature, Netdata's web log collector is unaware of these fields, but we can still get some info from them.
```

modules/weblog/README.md
```diff
@@ -261,6 +261,31 @@ we still can get some info from them.
             match: '* *'
 ```
 
+## Custom Time Fields Feature
```
joelhans, line 264:
```suggestion
## Custom time fields feature
```

modules/weblog/README.md
```diff
@@ -261,6 +261,31 @@ we still can get some info from them.
...
+
+Weblog is also able to extract user defined time fields and could count min/avg/max + histogram against these fields.
+
+This feature needs:
+-   custom log format with user defined time fields
```
joelhans, line 269:
```suggestion
-   A custom log format with user-defined time fields.
```

modules/weblog/README.md
```diff
@@ -261,6 +261,31 @@ we still can get some info from them.
...
+-   histogram to show response time in seconds, which is optional.
```
joelhans, line 270:
```suggestion
-   A histogram to show response time in seconds, which is optional.
```

modules/weblog/README.md
```diff
@@ -261,6 +261,31 @@ we still can get some info from them.
...
+
+As an example, apache [`mod_logio`](https://httpd.apache.org/docs/2.4/mod/mod_logio.html) adds `^FB` logging directive. This value shows a delay in microseconds between when the request arrived, and the first byte of the response headers are written.
```
joelhans, line 272:
```suggestion
As an example, Apache [`mod_logio`](https://httpd.apache.org/docs/2.4/mod/mod_logio.html) adds a `^FB` logging directive. This value shows a delay in microseconds between when the request arrived, and the first byte of the response headers are written.
```

vlvkobal (approved):

PR #486, opened 2020-11-11
Add module for monitoring file/folder size like Linux 'du' command

ski2per:
Signed-off-by: Ted <ski2per@gmail.com>

Hi~ We are using Netdata as our production monitoring system. It's really an awesome tool. 
In our production system, we have  requirement to monitor the size for some folders.
I have read Netdata docs and googled for some while, find no solutions.
So I try to write a go plugin for it, wish it would help :)

### DIFF
--- a/modules/du/du.go
+++ b/modules/du/du.go
@@ -1,0 +1,101 @@
+package du
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/netdata/go.d.plugin/agent/module"
+)
+
+// Config is configuration for Du module.
+type Config struct {
+	Paths []string `yaml:"paths"`
+}
+
+// Du is a module for file/folder size monitoring like linux command 'du'
+type Du struct {
+	module.Base // should be embedded by every module
+	Config      `yaml:",inline"`
+
+	charts        *module.Charts
+	collectedDims map[string]bool
+}
+
+func init() {
+	module.Register("du", module.Creator{
+		Defaults: module.Defaults{
+			UpdateEvery:        module.UpdateEvery,
+			AutoDetectionRetry: module.AutoDetectionRetry,
+			Priority:           module.Priority,
+			Disabled:           true,
+		},
+		Create: func() module.Module { return New() },
+	})
+}
+
+func fileSize(path string) (int64, error) {
+	var size int64
+	err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsDir() {
+			size += info.Size()
+		}
+		return err
+	})
+	return size, err
+}
+
+// New creates Du with default values.
+func New() *Du {
+	return &Du{
+		Config: Config{
+			Paths: make([]string, 0),
+		},
+		collectedDims: make(map[string]bool),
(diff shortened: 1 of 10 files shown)


modules/du/du.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+		Create: func() module.Module { return New() },
+	})
+}
+
+func fileSize(path string) (int64, error) {
+	var size int64
+	err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
```
thiagoftsm, line 40: Hello @ski2per ,

This function returns two values according its declaration, but here you are returning only one value, please, fix this.

modules/du/du.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+		if !info.IsDir() {
+			size += info.Size()
+		}
+		return err
```
thiagoftsm, line 45: The same think is happening here.

modules/du/du.go
thiagoftsm: sorry , I missed the fact this is inside another function.

ski2per: Wish it would work, i just jump from python to go 🤣

ilyam8:
Hi @ski2per 

`filecheck` collector monitors files size

> https://github.com/netdata/go.d.plugin/tree/master/modules/filecheck#filecheck

As for directories it monitors:
 - existence
 - time since the last modification
 - number of files

I guess we could extend it to monitor summarize disk usage. What do you think?

ski2per:
> Hi @ski2per
> 
> `filecheck` collector monitors files size
> 
> > https://github.com/netdata/go.d.plugin/tree/master/modules/filecheck#filecheck
> 
> As for directories it monitors:
> 
> * existence
> * time since the last modification
> * number of files
> 
> I guess we could extend it to monitor summarize disk usage. What do you think?



> Hi @ski2per
> 
> `filecheck` collector monitors files size
> 
> > https://github.com/netdata/go.d.plugin/tree/master/modules/filecheck#filecheck
> 
> As for directories it monitors:
> 
> * existence
> * time since the last modification
> * number of files
> 
> I guess we could extend it to monitor summarize disk usage. What do you think?

I think I might

> Hi @ski2per
> 
> `filecheck` collector monitors files size
> 
> > https://github.com/netdata/go.d.plugin/tree/master/modules/filecheck#filecheck
> 
> As for directories it monitors:
> 
> * existence
> * time since the last modification
> * number of files
> 
> I guess we could extend it to monitor summarize disk usage. What do you think?

I think I might overlook filecheck module. Extending it should be a good idea. Let me have a try.

I will close this PR

ski2per:
> Hi @ski2per
> 
> `filecheck` collector monitors files size
> 
> > https://github.com/netdata/go.d.plugin/tree/master/modules/filecheck#filecheck
> 
> As for directories it monitors:
> 
> * existence
> * time since the last modification
> * number of files
> 
> I guess we could extend it to monitor summarize disk usage. What do you think?

I think I might overlook the filecheck module. Extending it should be a good idea. Let me have a try.

I will close this PR now.

PR #487, opened 2020-11-12
Extend filecheck module with dir size chart

ski2per:
Signed-off-by: Ted <ski2per@163.com>

Hi, it's me again. I create this PR for this [#486 ](https://github.com/netdata/go.d.plugin/pull/486).
Just move the file/dir calculation func into filecheck module.


ilyam8:
Hi @ski2per 

i think we need to make dir size calculating optional (it could consume a lot of cpu if there are a lot of files) and likely use other walk dir library (ex.: https://github.com/karrick/godirwalk), but we can do it in another PR.

ilyam8:
And seems like we need an `recursive` option - `calDirNumOfFiles` returnes num of files in the root dir, `calcDirSize` does it recursively 🤔 On the other hand `numOfFiles` includes number of dirs, so it is kinda ok i think.

ilyam8 (approved):

modules/filecheck/collect_dirs.go
```diff
@@ -59,6 +63,10 @@ func (fc *Filecheck) addDirToCharts(dirpath string) {
 			id = dirDimID(dirpath, "mtime_ago")
 		case dirNumOfFilesChart.ID:
 			id = dirDimID(dirpath, "num_of_files")
+		case dirNumOfFilesChart.ID:
+			id = dirDimID(dirpath, "num_of_files")
```
vlvkobal, line 67: A duplicate. Please remove it.

ilyam8: hawk eye! 🐦

ilyam8, line 67:
```suggestion
```

ilyam8: removed in https://github.com/netdata/go.d.plugin/pull/487/commits/edc7b2f627bdd88d158468c04493ca3b13af7285

vlvkobal (approved):

ilyam8:
Many thanks @ski2per !

ski2per:
> Hi @ski2per
> 
> i think we need to make dir size calculating optional (it could consume a lot of cpu if there are a lot of files) and likely use other walk dir library (ex.: https://github.com/karrick/godirwalk), but we can do it in another PR.

I have thought this before, I will do it in another PR.

ski2per:
> Many thanks @ski2per !

You are welcome 😄

ilyam8:
i opened an issue https://github.com/netdata/go.d.plugin/issues/488

As for library don't hurry with the PR - i'll open an issue and we need to do benchmarks first.

PR #493, opened 2020-11-18
Change configure doc to reflect new standard

joelhans:
See netdata/netdata#10254 for an explanation. Fixing a grammatical issue and pointing users to the correct source of truth on Agent configuration.


ilyam8 (approved):

ilyam8:
Thanks!

Issue #500, opened 2020-11-19
`powerdns` migration:  authoritative server


ilyam8:
Python [`powerdns`](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/powerdns) collects data from powerdns Authoritative Server and Recursor.

Documentation: https://www.powerdns.com/documentation.html.

---

## Authoritative Server

- [Docker deployment](https://github.com/pschiffe/docker-pdns#pdns-mysql). Needs mysql/mariadb backend.
- [Documentation](https://doc.powerdns.com/authoritative/).
- [Enabling HTTP API/URL Ednpoints](https://doc.powerdns.com/authoritative/http-api/index.html#enabling-the-api).
- [Prometheus endpoint](https://doc.powerdns.com/authoritative/http-api/index.html#metrics-endpoint). (added in v4.4.0)
- [Statistics endpoint](https://doc.powerdns.com/authoritative/http-api/statistics.html).

❗ Statistics can be gathered via unix socket (`/var/run/pdns/pdns.controlsocket`), would be nice to add i think.

PR #501, opened 2020-11-19
add `powerdns` collector (authoritative ns)

ilyam8:
Fixes: #500

**A lot of room for improvements**:
- powerdns authoritative server exposes much more metrics then we collect
- our grouping could be better

However, i think that we need not to do it now, but for sure and certain we need to do it later.

So i use same charts as python collector has (there are some little changes still).


config/go.d/powerdns.conf
```diff
@@ -0,0 +1,168 @@
...
+#    Syntax:
+#      timeout: 1
+#
+#  - method
+#    HTTP request method.
+#    Syntax:
+#      method: GET
+#
+#  - body
+#    HTTP request method.
```
thiagoftsm, line 105: The `body` has the exact same description of `method`.

config/go.d/powerdns.conf
```diff
@@ -0,0 +1,168 @@
...
+#    Syntax:
+#      body: '{fake: data}'
+#
+#  - headers
+#    HTTP request headers.
+#    Syntax:
+#      headers:
+#        X-API-Key: key
```
thiagoftsm, line 113: Should we keep this header considering that we have `api_key`?

config/go.d/powerdns.conf
ilyam8: it is copy/paste, we have that typo in a lot of configs, i will fix it in another PR

config/go.d/powerdns.conf
ilyam8: It is an example of `headers` option usage.

Some APIs use API keys for authorization, i thought that we could add `x_api_key` or `api_key` option to the https://github.com/netdata/go.d.plugin/tree/master/pkg/web#configuration-options

```yaml
x_api_key: secret
```
is equal to

```yaml
headers:
   X-API-KEY: secret
```

but easier/faster to set

ilyam8: ok, i removed the option, will add it in a separate PR.

thiagoftsm (approved):

PR #503, opened 2020-11-21
add dnsmasq dns forwarder collector

ilyam8:
https://manpages.debian.org/stretch/dnsmasq-base/dnsmasq.8.en.html#NOTES

> The cache statistics are also available in the DNS as answers to queries of class CHAOS and type TXT in domain bind. The domain names are cachesize.bind, insertions.bind, evictions.bind, misses.bind, hits.bind, auth.bind and servers.bind. An example command to query this, using the dig utility would be

### DIFF
--- a/config/go.d/dnsmasq.conf
+++ b/config/go.d/dnsmasq.conf
@@ -61,10 +61,10 @@
 #    Syntax:
 #      address: 127.0.0.1:53
 #
-#  - network
+#  - protocol
 #    DNS query transport protocol. Valid options: udp, tcp, tcp-tls.
 #    Syntax:
-#      network: udp
+#      protocol: udp
 #
 #  - timeout
 #    DNS query timeout (dial, write and read) in seconds.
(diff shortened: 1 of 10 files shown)


config/go.d/dnsmasq.conf
```diff
@@ -0,0 +1,100 @@
...
+#     param1: value1
+#
+#
+# [ List of JOB specific parameters ]:
+#  - address
+#    Server's address. Format is 'ip_address:port'.
+#    Syntax:
+#      address: 127.0.0.1:53
+#
+#  - network
```
thiagoftsm, line 64: Why are we using `network` instead `protocol` here? I think this can create confusion with `network address`.

ilyam8: dns lib client uses network (net)

```go
// A Client defines parameters for a DNS client.
type Client struct {
	Net       string      // if "tcp" or "tcp-tls" (DNS over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
...
```

ilyam8: I am ok with `protocol` too.

thiagoftsm (approved):

PR #506, opened 2020-11-24
modules/mysql: simplify version parsing

ilyam8:
ssia


ilyam8:
just variables renaming basically, no reviews needed

PR #509, opened 2020-11-25
dnsdist: fix data type

thiagoftsm:
While I was testing the PR https://github.com/netdata/netdata/pull/10284 I observed the following error:

```
 ERROR: dnsdist[local] error on decoding response from 'http://127.0.0.1:8083/jsonstat?command=stats': json: cannot unmarshal number 12892.215313625313 into Go struct field statisticMetrics.latency-avg100 of type int64 
```

This PR fixes the bug.


ilyam8:
I guess there are not only floats. And we need to update testdata

https://github.com/netdata/go.d.plugin/blob/master/modules/dnsdist/testdata/v1.5.1/jsonstat.json

modules/dnsdist/metrics.go
```diff
@@ -4,36 +4,36 @@ package dnsdist
...
+	CPUSysMsec            int64   `stm:"cpu-sys-msec" json:"cpu-sys-msec"`
+	CPUUserMsec           int64   `stm:"cpu-user-msec" json:"cpu-user-msec"`
+	DownStreamSendErrors  int64   `stm:"downstream-send-errors" json:"downstream-send-errors"`
+	DownStreamTimeout     int64   `stm:"downstream-timeouts" json:"downstream-timeouts"`
+	DynBlocked            int64   `stm:"dyn-blocked" json:"dyn-blocked"`
+	EmptyQueries          int64   `stm:"empty-queries" json:"empty-queries"`
+	LatencyAvg100         float64 `stm:"latency-avg100" json:"latency-avg100"`
+	LatencyAvg1000        float64 `stm:"latency-avg1000" json:"latency-avg1000"`
+	LatencyAvg10000       float64 `stm:"latency-avg10000" json:"latency-avg10000"`
+	LatencyAvg1000000     float64 `stm:"latency-avg1000000" json:"latency-avg1000000"`
```
ilyam8, line 19: testdata should be updated so we have floats in there

ilyam8 (changes requested):

modules/dnsdist/metrics.go
thiagoftsm: I updated.

ilyam8 (approved):

PR #513, opened 2020-11-29
Couchbase plugin

hamedbrd:
**Summary**

A simple Couchbase module to show basic metrics per bucket netdata charts, tests and README is not added yet.

Fixes [#4811](https://github.com/netdata/netdata/issues/4811)


ilyam8:
@hamedbrd hi, thx for the PR 👍 

i have a second thought about this collector, i think we don't need to use any library since it is a couchbase has [well-documented REST API](https://developer.couchbase.com/resources/best-practice-guides/monitoring-guide.pdf).

We are using `/pools/default/buckets` endpoint in the initial version, but it is likely we will be using more.

My suggestion is to use [elasticsearch](https://github.com/netdata/go.d.plugin/tree/master/modules/elasticsearch) module as an example and rewrite the module using std http package.

hamedbrd:
@ilyam8 how about chart types? I'm not sure their types are correct or not!

ilyam8:
We chose 1st option from https://github.com/netdata/netdata/issues/4811#issuecomment-727877665, so i think the type is `module.Statcked` for all the charts!

hamedbrd:
@ilyam8 may you have a quick look and let me know what you think? tests and documentation are not done yet

ilyam8:
@hamedbrd ofc, will take a look later today, thx for keep working on it!

modules/couchbase/collect.go
```diff
@@ -0,0 +1,183 @@
...
+		collected[indexDimID(b.Name, "disk")] = int64(bs.DiskUsed)
+		collected[indexDimID(b.Name, "data")] = int64(bs.DataUsed)
+		collected[indexDimID(b.Name, "mem")] = int64(bs.MemUsed)
+		collected[indexDimID(b.Name, "num_non_resident")] = int64(bs.VbActiveNumNonResident)
+	}
+
+	return nil
+}
+
+func (cb *Couchbase) addDimsToCharts(bucket string) error {
```
ilyam8, line 55: Check, i think it is easier to follow https://github.com/netdata/go.d.plugin/blob/d2b0731a19a22e11fe450329729bbe9ba9a37702/modules/redis/collect_info.go#L166-L194

modules/couchbase/collect.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil, fmt.Errorf("error on creating a connection: %v", err)
+	}
+
+	return collected, nil
+}
+
+func (cb Couchbase) collectBasicStats(collected map[string]int64, ms *cbMetrics) error {
+	for _, b := range ms.BucketsStats {
+
+		if !cb.collectedBuckets[b.Name] && ms.hasBucketsStats() {
```
ilyam8, line 36: > ms.hasBucketsStats()

is not needed, b/c we `range ms.BucketsStats`, there is no loop if len(ms.BucketsStats) < 1

modules/couchbase/collect.go
```diff
@@ -0,0 +1,183 @@
...
+func (cb Couchbase) scrapeBucketsStats(ms *cbMetrics) {
+	req, _ := web.NewHTTPRequest(cb.Request)
+	req.URL.Path = urlPathBucketsStats
+
+	var stats []bucketsStats
+	if err := cb.doOKDecode(req, &stats); err != nil {
+		cb.Warning(err)
+		return
+	}
+	for _, bucket := range stats {
```
ilyam8, line 144: i think `ms.BucketsStats = stats` will do, range loop is not needed.

Also `if err := cb.doOKDecode(req, &ms.BucketsStats); err != nil` should work also

modules/couchbase/charts.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"github.com/netdata/go.d.plugin/agent/module"
+)
+
+type (
+	Charts = module.Charts
+	Chart  = module.Chart
+	Dim    = module.Dim
+)
+
```
ilyam8, line 12: i suggest to create a var `bucketCharts`, so you can

```go
   return bucketCharts.Copy(), nil
```

in `initCharts()`

Like

```go
var bucketCharts = module.Charts{
	dbPercentCharts.Copy(),
    ...
}
```

modules/couchbase/charts.go
```diff
@@ -0,0 +1,76 @@
...
+var dbPercentCharts = Chart{
+	ID:    "couchbase_quota_percent_used_stats",
+	Title: "Quota Percent Used Per Bucket",
+	Units: "%",
+	Fam:   "quota percent used",
+	Ctx:   "couchbase.couchbase_quota_percent_used_stats",
```
ilyam8, line 18: lets make it `couchbase.couchbase_quota_percent_used_stats` => `couchbase.quota_percent_used_stats` (revmoe `couchbase` after `.`)

Please do it for all contextes

ilyam8:
@hamedbrd there are some things to fix, but looks pretty good to me!

> tests and README is not added yet.

Lets add tests, it would be much easier for you to make changes if you have them

ilyam8:
@hamedbrd ops something went wrong, if you wont be able to resolve it feel free to close the PR and open new one 😉

hamedbrd:
@ilyam8 yes, I just rebased the branch
I do not know what happened

PR #524, opened 2020-12-08
Energid

### LINKED ISSUE #512: `energid` migration
We are working to migrate the `energid` collector to golang. To do the migration the developer can get it direct from [here](https://docs.energi.software/en/downloads/core-node).

For more details about how to use the binary, we have [this documentation](https://github.com/energicryptocurrency/energi3/blob/master/README.md).

And to have some request outputs, we can use [this link](https://github.com/energicryptocurrency/core-api-documentation)

thiagoftsm:
Fixes #512 

More details after tests I am working.


ilyam8:
As we discussed we need to change several concurrent http requests to just one, it is possible to query all metrics in one request.

modules/energid/collect.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+func (e *Energid) collect() (map[string]int64, error) {
+	ms := e.scrapeEnergid()
+
+	if ms == nil {
+		return nil, fmt.Errorf("Invalid request")
+	}
+
+	collected := make(map[string]int64)
```
ilyam8, line 26: i think you just need to set `stm` tags you want and then stm.ToMap energidStats and not specific parts of it

modules/energid/charts.go
```diff
@@ -0,0 +1,108 @@
...
+			Title: "Secure Memory",
+			Units: "KiB",
+			Fam:   "memory",
+			Ctx:   "energid.secmem",
+			Type:  module.Area,
+			Dims: module.Dims{
+				{ID: "secmem_total", Name: "Total", Algo: module.Absolute},
+				{ID: "secmem_locked", Name: "Locked", Algo: module.Absolute},
+				{ID: "secmem_used", Name: "Used", Algo: module.Absolute},
+			},
```
ilyam8, line 61: respone has no `secmem_*`? if so just delete it

ilyam8 (changes requested):

modules/energid/charts.go
thiagoftsm: We do not have it more running on `2.4.x`, I believe it was removed. I removed it.

modules/energid/energid_test.go
```diff
@@ -0,0 +1,243 @@
...
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/netdata/go.d.plugin/pkg/tlscfg"
+	"github.com/netdata/go.d.plugin/pkg/web"
+	"github.com/prometheus/common/log"
```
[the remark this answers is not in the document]
thiagoftsm: You are right, `VScode` adds automatic the libraries, and this was added by it :/ , I set the correct library. Thanks!

modules/energid/collect.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	if err := e.doOKDecode(req, &stats); err != nil {
+		e.Warning(err)
+		return nil
+	}
+
+	return ms
+}
+
+func (e *Energid) energidMakeRequest() *energyRequests {
```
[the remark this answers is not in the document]
thiagoftsm: I changed to `var`.

ilyam8 (approved):

vlvkobal (approved):

PR #525, opened 2020-12-09
add glob support for filecheck modules directories

BT-dschleich:
I tried to add glob support for the filecheck modules directory check, as this is a great extension of usability to monitor dynamic directories.
So far it worked on my tests I used in our test instance.

I hope the code is "ok" with my so far limited go experience.

### DIFF
--- a/modules/filecheck/README.md
+++ b/modules/filecheck/README.md
@@ -64,7 +64,7 @@ cd /etc/netdata # Replace this path with your Netdata config directory
 sudo ./edit-config go.d/filecheck.conf
 ```
 
-Needs only a path to a file or a directory. **The path doesn't support any wildcards**.
+Needs only a path to a file or a directory. Wildcards are supported.
 
 Here is an example:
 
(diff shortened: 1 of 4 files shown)


ilyam8:
@BT-dschleich hi, thx for the contribution!

>  as this is a great extension of usability to monitor dynamic directories.

Couldnt agree more! Glob support is must have for this collector. However, it wasn't implemented from the beginning, because it is not easy as it looks. And we need to discuss it first.

For instance:
 - we don't want to show files/dirs patterns (has meta) as dimensions in charts (they are dimensions in your impl if `filepath.Glob` returns nil).
 - perhaps we don't want to call `filepath.Glob` on every `Collect`, but implement it as a discovery phase which runs every `discovery_every` or something.
 - likely we want to add some cleanup for discovered files/dirs (pattern with meta)

I think we need an issue and discuss implementation in there, agree on it, then implement. What do you think?

BT-dschleich:
@ilyam8 hi and thanks for your feedback.

Sure lets continue this in an issue, if this is the most productive way to implement it properly. 

My go knowledge sadly is quite limited. 
Do you have an example of a module implementing such a discovery mechanism?

PR #527, opened 2020-12-11
docs: auto format md files, make format consistent in this project

ilyam8:
ssia


ilyam8:
Netdata docs md format uses 3 spaces after `-` which i don't like, i would prefer to have default 1 space for lists, i think it is ok to not follow netdata project in that aspect, i will change to 3 if there is strong reason to do so.

vlvkobal (approved):

ilyam8:
If there are any problems (likely) i will change them in a separate PR, going to spend some time on it on weekends.

PR #515, opened 2020-11-30
Nginxvts plugin

ski2per:
Hi,
I am trying to add NginxVts support for go.d.plugin. It supports most of NginxVts metrics currently.
Also, wish you guys can help me with this :)

PR #543, opened 2021-01-21
modules/nginx-vts: tune charts and small cleanup

ilyam8:
@ski2per


ski2per (commented):
this is really helpful :)

ilyam8:
@ski2per from your experience - is having not only total, but per server/upstream/filter metrics super important? I assume the answer is yeas, ofc. 

Lets put it this way - is having only summary provides enough value or w/o per server/upstream/filter metrics it is not really usable?

ski2per:
> @ski2per from your experience - is having not only total, but per server/upstream/filter metrics super important? I assume the answer is yeas, ofc.
> 
> Lets put it this way - is having only summary provides enough value or w/o per server/upstream/filter metrics it is not really usable?

@ilyam8 ,  IMO if the nginx only serves one virtualhost like www.demo.com, this total statistics should be enough. But in our production environment, we have (a | b | c ...) .demo.com virtualhost served on one Nginx.
So I forked another go.d.plugin to fill these stuffs(server_name, upstreams, filter) into chart and family fields of dims, then use regexp to extract this stuffs on Grafana page. Kind of weird, but will do the work. 
As Nginx VTS configurations is very dynamic (especially for the "filter", it can use nginx variables as its value), I think add labels to chart or dim will be a good solution to this. (as you mentioned this in [#515 ](https://github.com/netdata/go.d.plugin/pull/515#issuecomment-735753404)

PR #552, opened 2021-02-15
add supervisord module

ilyam8:
Fixes: netdata/netdata#4852

This PR adds initial [Supervisor](http://supervisord.org/) monitoring.

Collected metrics:

- Summary
  - Running/non-running processes
- [Per Process' group](http://supervisord.org/api.html#supervisor.rpcinterface.SupervisorNamespaceRPCInterface.getProcessInfo)
  - Running/non-running processes
  - State code (per process)
  - Exit status (per process)
  - Uptime (per process)
  - Downtime (per process)

@ekexcello i decided to add [all the summary charts](https://github.com/netdata/netdata/issues/4852#issuecomment-758039791) in the 2nd PR, i think we need to discuss them a little bit more.

### DIFF
--- a/modules/supervisord/collect.go
+++ b/modules/supervisord/collect.go
@@ -1,0 +1,137 @@
+package supervisord
+
+import (
+	"fmt"
+
+	"github.com/netdata/go.d.plugin/agent/module"
+)
+
+const (
+	// http://supervisord.org/subprocess.html#process-states
+	// STOPPED  (0)
+	// STARTING (10)
+	// RUNNING (20)
+	// BACKOFF (30)
+	// STOPPING (40)
+	// EXITED (100)
+	// FATAL (200)
+	// UNKNOWN (1000)
+
+	stateRunning = 20
+)
+
+func (s *Supervisord) collect() (map[string]int64, error) {
+	info, err := s.client.getAllProcessInfo()
+	if err != nil {
+		return nil, err
+	}
+
+	ms := make(map[string]int64)
+	s.collectAllProcessInfo(ms, info)
+
+	return ms, nil
+}
+
+func (s *Supervisord) collectAllProcessInfo(ms map[string]int64, info []processStatus) {
+	ms["running_processes"] = 0
+	ms["non_running_processes"] = 0
+	for _, p := range info {
+		if !s.collectedGroups[p.group] {
+			s.collectedGroups[p.group] = true
+			s.addProcessGroupCharts(p)
+		}
+		id := procID(p)
+		if !s.collectedProcesses[id] {
+			s.collectedProcesses[id] = true
+			s.addProcessToCharts(p)
+		}
+
+		ms[fmt.Sprintf("group_%s_running_processes", p.group)] += 0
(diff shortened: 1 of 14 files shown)


modules/supervisord/collect.go
```diff
@@ -0,0 +1,172 @@
...
+	// http://supervisord.org/subprocess.html#process-states
+	// STOPPED  (0)
+	// STARTING (10)
+	// RUNNING (20)
+	// BACKOFF (30)
+	// STOPPING (40)
+	// EXITED (100)
+	// FATAL (200)
+	// UNKNOWN (1000)
+	return p.state == 20
```
ilyam8, line 171: Seems this is not correct, should be `STARTING || RUNNING || STOPPING` 🤔

ilyam8: @ekexcello what do you think? Uptime [`STARTING`, `STOPPING`], downtime [`STOPPING`, `STARTING`]

vlvkobal (approved):

modules/supervisord/collect.go
ekexcello: Hello @ilyam8 
Sorry, but looking at function code I can't get the question, could you please rephrase it?

ilyam8: Do we consider process as running if its state is `STARTING`/`STOPPING`?
We need this to have begin/end points to calc uptime/downtime.

ilyam8: I think yes. And i need to update the code 😄

ilyam8: On the other hangs, if the process is hanging in `STARTING`.. We don't want to treat it as running and show its uptime, it is still downtime 🤔

ekexcello: I agree that process in "starting" state is not yet running, and "stopping" is for sure not up. However, I'll also ask opinion of my colleagues, will update you later providing their opinion, too

ilyam8: Keep in mind that visualising `state code` and `exit status` until we don't do averaging - only for real-time data. Averaging is ok for some metrics (number of running processes) and doesn't work at all for other (e.g. `state code`)

That is the data for the last 30 minutes (actual state codes values: 20, 100).

Screenshot 2021-02-16 at 16 25 58

ekexcello: Charts look promising! However, it would be good to add explanation with state code values if possible.

> I agree that process in "starting" state is not yet running, and "stopping" is for sure not up. 
Opinion of my colleagues is very same: "starting means not running".

ilyam8: > However, it would be good to add explanation with state code values if possible.

Yeah, i will add it. I need to update `dashboard_info.js` file in the netdata repo...

---

> stopping means running - it's not yet stopped

ok, so lets open new issue (i will do it) and discuss all fixes/improvements in there. My goal was to provide minimal requested functionality and improve later when we have it refined.

ekexcello:
Thank you very much, @ilyam8 !

PR #559, opened 2021-03-11
Support VS Code container deven

OdysLam:
Summary
With the existence of the .devcontainer directory, when a user opens the repository using VS Code, they will be automatically prompted to re-open the directory from inside the developer container. T

The VS Code window is connected to that container, being able to both use extensions to facilitate programming, tools. to check that the code is according to our standards (e.g flake8) and use VS Code sidebar as a GUI for file navigation inside the container. Finally, VS Code syncs the directory that we just opened inside the container, meaning that any file we place from our host machine into that directory (e.g netdata), it will be synced inside the container.

The developer container contains all the required dependencies to build from source:

- netdata
- python.d.plugin
- go.d.plugin
- netdata dashboard

The new contribution workflow will be the following:

- Fork the netdata/go.d.plugin repository
- git clone the forked repository locally
- open the directory using VS Code
- reopen from inside the devenv, ready to develop!

This new optional workflow greatly simplifies the contribution process, since the user only needs to have VS Code installed.

Component Name
Test Plan
Download the repository and open the directory using VS Code

Additional Information
- https://code.visualstudio.com/docs/remote/containers
- https://github.com/netdata/community/tree/main/devenv

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,14 @@
+{
+    "image": "netdata/devenv", 
+    "extensions":[
+		"golang.go",
+		"exiasr.hadolint",
+		"ms-python.python",
+		"timonwong.shellcheck",
+		"redhat.vscode-yaml",
+		"dbaeumer.vscode-eslint",
+		"jasonnutter.search-node-modules",
+		"mgmcdermott.vscode-language-babel",		
+	],
+    "forwardPorts": [19999]
+}


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,14 @@
...
+    "image": "netdata/devenv", 
+    "extensions":[
+		"golang.go",
+		"exiasr.hadolint",
+		"ms-python.python",
+		"timonwong.shellcheck",
+		"redhat.vscode-yaml",
+		"dbaeumer.vscode-eslint",
+		"jasonnutter.search-node-modules",
+		"mgmcdermott.vscode-language-babel",
```
ilyam8, line 11: Do we need all those extension to develop in golang?

OdysLam: You are correct. Removing them.

OdysLam: Actually, we have a single devenv for all of Netdata, so I would argue that the overhead is minimal.  I would prefer to have the same support for every repo.

ilyam8: So if you want to add an extensions for python you are going to update all the repositories?

OdysLam: That's the idea. What do you think @ilyam8 ? Perhaps @Ferroin could give his view as well. I just want the best possible contributor experience (less friction as possible)

ilyam8 (approved):

ilyam8:
I don't get why those files should be in sync between repositories. But it is ok. We can remove unrelated extensions later if this is a problem.

PR #581, opened 2021-06-07
add caddy prometheus endpoint

OdysLam:


config/go.d/prometheus.conf
```diff
@@ -1518,3 +1518,5 @@ jobs:
     url: 'http://127.0.0.1:44323/metrics'
   - name: dcos_exporter_local
     url: 'http://127.0.0.1:61091/metrics'
+  - name: caddy_local
+    url: 'http://localhost:2019/metrics'
```
ilyam8, line 1522:
```suggestion
  - name: caddy_local
    url: 'http://localhost:2019/metrics'
    expected_prefix: 'caddy_'
```

I suggest adding [that option](https://github.com/netdata/go.d.plugin/blob/4e1609345572c80362a27b57c94829364c93d320/config/go.d/prometheus.conf#L104-L108).

This is needed to ensure that `:2019` app is Caddy.

ilyam8 (changes requested):

ilyam8 (approved):

PR #583, opened 2021-06-14
vsphere - datastores

Valamyrr:
Making vSphere module for Netdata's Go plugin able to monitor vsphere datastores (it was already able to monitor hosts and vms).
There is also a little modification on error handling in collect.go : before, an error prevented any metric to be collected, so i edited it so that, for example, hosts and datastores metrics can be collected even when vms metrics cannot be collected.

### DIFF
--- a/modules/vsphere/vsphere_test.go
+++ b/modules/vsphere/vsphere_test.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 	"testing"
 	"time"
-	"regexp"
 
 	"github.com/netdata/go.d.plugin/modules/vsphere/discover"
 	"github.com/netdata/go.d.plugin/modules/vsphere/match"
@@ -300,8 +299,7 @@ func TestVSphere_Collect(t *testing.T) {
 
 	//working around ioutil.TempDir naming convention
 	for key, value := range collected {
-		match,_ := regexp.MatchString(".*LocalDS_.*",key)
-		if(match){
+		if(strings.Contains(key,"LocalDS_")){
 			expected[key]=value
 		}
 	}
--- a/modules/vsphere/vsphere_test.go
+++ b/modules/vsphere/vsphere_test.go
@@ -86,6 +86,9 @@ func TestVSphere_Init_ReturnsFalseIfInvalidHostVMIncludeFormat(t *testing.T) {
 
 	vSphere.VMsInclude = match.VMIncludes{"invalid"}
 	assert.False(t, vSphere.Init())
+
+	vSphere.DatastoresInclude = match.DatastoreIncludes{"invalid"}
+	assert.False(t, vSphere.Init())
 }
 
 func TestVSphere_Check(t *testing.T) {
@@ -297,8 +300,9 @@ func TestVSphere_Collect(t *testing.T) {
 	count := model.Count()
 	assert.Len(t, vSphere.discoveredHosts, count.Host)
 	assert.Len(t, vSphere.discoveredVMs, count.Machine)
-	assert.Len(t, vSphere.charted, count.Host+count.Machine)
-	assert.Len(t, *vSphere.charts, count.Host*len(hostCharts)+count.Machine*len(vmCharts))
+	assert.Len(t, vSphere.discoveredDatastores, count.Datastore)
+	assert.Len(t, vSphere.charted, count.Host+count.Machine+count.Datastore)
+	assert.Len(t, *vSphere.charts, count.Host*len(hostCharts)+count.Machine*len(vmCharts)+count.Datastore*len(datastoreCharts))
 	ensureCollectedHasAllChartsDimsVarsIDs(t, vSphere, collected)
 }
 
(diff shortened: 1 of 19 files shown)


ilyam8:
@Valamyrr hi, thanks for the contribution.

The PR doesn't compile with the following error

> modules/vsphere/collect.go:182:4: vs.updateDatastoresCharts undefined (type *VSphere has no field or method updateDatastoresCharts)

---

> Making vSphere module for Netdata's Go plugin able to monitor vsphere datastores (it was already able to monitor hosts and vms)

Yeah, I didn't add datastore metrics, because they are not real-time (which are available only for hosts and virtual machines). And historical available at a (default) 5 minute, 30 minutes, 2 hours, and 24 hours rollup levels. Likely there were other reasons, but I don't remember 😄 

Having said that, I agree that we need to collect datastore metrics.

---

> for example, hosts and datastores metrics can be collected even when vms metrics cannot be collected.

Is that possible? I guess it is either we can collect everything or nothing.

Valamyrr:
Hi @ilyam8 !
Yes, I forgot some things, I am working on it.
I made this a while ago (before the other PR actually), so I don't remember all the details either...

> Is that possible? I guess it is either we can collect everything or nothing.
The way it was handled is to return any error during the collection. I just returned all the errors (they were strings) at once by concatenating them, so that the collection can resume with the other objects. Maybe it's a bad behavior, you can see it in the 20-30 first lines of collect.go. It probably is (I am 100% sure there is a better way to do it than what I did). But I think there was an error when no vms metrics were collected (all vms were filtred for example) so it could not pursue with datastores metrics and it was not good for how I intended to use the module.

Valamyrr:
Hi @ilyam8 ,
I meet some troubles when I try to increase the test coverage (which I didn't do earlier, the tests were passing and now they are not).
A mock datastore is scraped, but its ID is always changing (something alike `/tmp/govcsim-DC0-LocalDS_0-123456789@folder-5`).
Perhaps do you have a solution to this issue? (I never used govcsim to mock data before).

ilyam8:
@Valamyrr it's been a while, I don't remember the details, let me check it.

ilyam8:
Ok, I see the simulator uses [ioutil.TempDir](https://golang.org/pkg/io/ioutil/#TempDir) when creating datastores.

https://github.com/vmware/govmomi/blob/93851bd6e14df055d86661de494119e6bacb2192/simulator/model.go#L783-L790
=>
https://github.com/vmware/govmomi/blob/93851bd6e14df055d86661de494119e6bacb2192/simulator/model.go#L813-L833

So there is no way we have consistent names during testing.

LocalDS_0-123456789: `LocalDS_<index number>-<a random string>`

ilyam8:
> But I think there was an error when no vms metrics were collected (all vms were filtred for example)

I think we shouldn't treat it as an error. If we do it then it is a bug 🤔

ilyam8:
> But I think there was an error when no vms metrics were collected (all vms were filtred for example)

Oh, I see, we return an error if ScrapeVMs/ScrapeHosts return zero performance entity metrics

https://github.com/netdata/go.d.plugin/blob/70a519ed0ba57470cdb890d402aee6496274b2c6/modules/vsphere/collect.go#L100-L105

https://github.com/netdata/go.d.plugin/blob/70a519ed0ba57470cdb890d402aee6496274b2c6/modules/vsphere/collect.go#L37-L42

I think we shouldn't call the functions if `vs.resources.VMs`/`vs.resources.Hosts` are empty (not discovered).

Valamyrr:
> I think we shouldn't call the functions if vs.resources.VMs/vs.resources.Hosts are empty (not discovered).

I agree.

> So there is no way we have consistent names during testing.

Thank you for the info! Then, should I drop the datastores in the testing functions that fail because of the naming?

ilyam8:
> Then, should I drop the datastores in the testing functions that fail because of the naming?

I guess we can do a quick workaround and just add datastore related metrics to the `expected` from the `collected` 😉

Valamyrr:
You're right, so I did it just now. I will push everything (there are other tests that don't cover datastores, I think) in the end of the week.

Valamyrr:
I still have an issue with hierarchy and the workaround to collect DS metrics doesn't work as expected (I know why). Working on it.

modules/vsphere/vsphere_test.go
```diff
@@ -300,7 +300,7 @@ func TestVSphere_Collect(t *testing.T) {
 
 	//working around ioutil.TempDir naming convention
 	for key, value := range collected {
-		match,_ := regexp.MatchString("*LocalDS_*",key)
+		match,_ := regexp.MatchString(".*LocalDS_.*",key)
```
ilyam8, line 303: no need in using regexp, [`strings.Contains`](https://golang.org/pkg/strings/#Contains) will doo

Valamyrr:
Indeed. I was searching for a more accurate regex, but considering what you said earlier, it should be enough.
However, I struggle quite a bit with the documentation and I have a hard time with the hierarchy. What upsets me more is that it does work when I remove the datastores' hierarchy from everywhere ; the metrics are still collected wonderfully. I don't understand... Nevertheless, I'll give it more tries.

Valamyrr:
Apparently, I can't get the ParentID from Datastore.Parent.Value (in build.go), so I can't get the hierarchy together for datastores.
Whenever I try, the returned value is nil and the build panics through `runtime error: invalid memory address or nil pointer dereference`.
I don't use datastoreFolders thought, but it was not needed until then. Perhaps I should try.

ilyam8:
closing due to no activity

PR #598, opened 2021-07-25
mongodb collector [serverStatus only]

georgeok:
A mongo db collector. Currently it supports dbstat and server status. This is very much work in progress yet. Sharding and replication are missing. Charts need to be refined to appropriate types(stack, line, etc.)


ilyam8:
@georgeok there are some merge conflicts, it is not that important to fix them asap, but please do.

modules/mongodb/mongodb.go
```diff
@@ -0,0 +1,171 @@
...
+	case m.Local.valid():
+		client, err = mongo.NewClient(options.Client().ApplyURI(m.Local.connectionString()))
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), m.Timeout*time.Second)
+	defer cancel()
+	err = client.Connect(ctx)
```
ilyam8, line 131: There is a built-in retry feature, but we never retry if Init() fails. It is supposed to initialize config/internal structs only, not the connection (see https://github.com/netdata/go.d.plugin/pull/577).

ilyam8:
@georgeok could you add testdata? (example: [mysql](https://github.com/netdata/go.d.plugin/tree/master/modules/mysql/testdata), [redis](https://github.com/netdata/go.d.plugin/tree/master/modules/redis/testdata))

modules/mongodb/mongodb.go
georgeok: done

georgeok:
I have refactored the code to follow the same ideas like the other modules, added test data and tested it. I have a couple of things to do more and we will be good to go:

- [ ] Add ops latency charts 
- [ ] Correct the readme file with the appropriate info

modules/mongodb/testdata/v5.0.0/serverStatus.go
```diff
@@ -0,0 +1,2255 @@
+package v5_0_0
+
+var V500ServerStatus = `
```
ilyam8, line 3: Wow, that is a lot of data. Is it possible to query specific stats? Or it is either all or nothing?

ilyam8:
@georgeok don't forget we need to fix branch conflicts

modules/mongodb/testdata/v5.0.0/serverStatus.go
georgeok: We can only disable `repl`, `metrics `, `locks` in command or maybe use aggregation to filter out unwanted data but this will transfer the load from the collector to the database itself. I am not 100% what is the correct balance, some extra network IO or some CPU power from the database itself.

ilyam8: It is 2255 lines, how many if we disable those three? Or how many each of them adds?

georgeok: locks is about 38 lines of pretified json/bson and metrics is about 1189 lines of pretty json/bson or **685 values** and repl about 34 lines. Keep in mind that we do not collect them all. Just a portion.

georgeok: To add some more details, this big response is only for self deployed databases. Atlas reports quite a few less (150 lines of prettified json/bson) data and we product only 8 charts in this case. The other charts are only enabled if we get data back from the server at the appropriate sections.

ilyam8:
Nice work @georgeok, here is some suggestion based on [the README](https://github.com/georgeok/go.d.plugin/blob/mongodb-module/modules/mongodb/README.md#mongodb-monitoring-with-netdata):
 - let's rename `connectionStr` to `dsn` (or `uri`). We use `dsn` in the [`mysql`](https://github.com/netdata/go.d.plugin/tree/master/modules/mysql#configuration) module.
 - let's do not support python way configuration, but only `dsn`.
 - keep in mind that all charts **units** with `incremental` dimensions should have `/s` prefix ([`incremental`](https://learn.netdata.cloud/docs/agent/collectors/plugins.d#dimension) dimension algorithm means calculate per second rate and store that calculated value in the db).

georgeok:
@ilyam8 Done, I've renamed connectionStr to uri (suits better than dsn for mongo) and also added the /s suffix to the units of charts which have incremental dims and are not reporting time values.

modules/mongodb/config.go
```diff
@@ -0,0 +1,15 @@
+package mongo
```
ilyam8, line 1: My suggestion is to move Config to the `mongodb.go`. Why do you think having it in a separate file is better? (after removing python-config-way code)

modules/mongodb/mongodb.go
```diff
@@ -0,0 +1,114 @@
...
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+func init() {
+	module.Register("mongodb", module.Creator{
+		Defaults: module.Defaults{
+			UpdateEvery:        module.UpdateEvery,
+			AutoDetectionRetry: module.AutoDetectionRetry,
+			Priority:           module.Priority,
+			Disabled:           false,
```
ilyam8, line 18: These are defaults. Let's do not set them explicitly.

modules/mongodb/mongodb.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+		Create: func() module.Module { return New() },
+	})
+}
+
+func New() *Mongo {
+	return &Mongo{
+		Config: Config{
+			Timeout: defaultTimeout,
+			Uri:     defaultUri,
```
ilyam8, line 28: My suggestion is to remove both constants and use string values instead. In that case, you don't need to test these config options in TestNew. Whatever is set in New is valid.

modules/mongodb/mongodb.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (m *Mongo) Charts() *module.Charts {
+	return m.charts
+}
+
+func (m *Mongo) Collect() map[string]int64 {
+	if m.client == nil {
+		var err error
+		err = m.initMongoClient()
+		if err != nil {
```
ilyam8, line 69:
```suggestion
		if err := m.initMongoClient(); err != nil {
```

How about simplifying it?  We don't use `err` later in the code.

modules/mongodb/mongodb.go
```diff
@@ -0,0 +1,114 @@
...
+		return err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), m.Timeout*time.Second)
+	defer cancel()
+	err = client.Connect(ctx)
+	if err != nil {
+		return err
+	}
+	m.client = client
+	return nil
```
ilyam8, line 109: Do we need to call Ping here?

From [the documentation](https://github.com/mongodb/mongo-go-driver#usage)

> Calling Connect does not block for server discovery. If you wish to know if a MongoDB server has been found and connected to, use the Ping method

modules/mongodb/mongodb_test.go
```diff
@@ -0,0 +1,87 @@
+package mongo
+
+import (
+	"testing"
+	"time"
+
+	"github.com/netdata/go.d.plugin/agent/module"
+	"github.com/stretchr/testify/assert"
```
ilyam8, line 8: An empty line, please

modules/mongodb/mongodb_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestNew(t *testing.T) {
+	job := New()
+	assert.Implements(t, (*module.Module)(nil), job)
+	assert.Equal(t, time.Duration(defaultTimeout), job.Timeout, "wrong timeout")
+	assert.Equal(t, defaultUri, job.Uri, "wrong timeout")
```
ilyam8, line 16: No need to test that if we remove both default* constants.

modules/mongodb/testdata/v5.0.0/serverStatus.go
```diff
@@ -0,0 +1,2274 @@
+package v5_0_0
+
```
ilyam8, line 2: It looks like we can remove `v500` prefix from all variables in this package.

`v5_0_0.V500ServerStatus` => `5_0_0.ServerStatus`

What do you think?

modules/mongodb/collect_test.go
```diff
@@ -0,0 +1,116 @@
+package mongo
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/netdata/go.d.plugin/agent/module"
+	v5_0_0 "github.com/netdata/go.d.plugin/modules/mongodb/testdata/v5.0.0"
```
ilyam8, line 8: The package name is `v5_0_0`, need for an alias here.

README.md
```diff
@@ -119,7 +120,7 @@ forums. We have a whole [category](https://community.netdata.cloud/c/agent-devel
   locally the **forked** repository (e.g `git clone https://github.com/odyslam/go.d.plugin`).
 - Using a terminal, `cd` into the directory (e.g `cd go.d.plugin`)
 - Add your module to the [modules](https://github.com/netdata/go.d.plugin/tree/master/modules) directory.
-- Import the module in the [main.go](https://github.com/netdata/go.d.plugin/blob/master/cmd/godplugin/main.go).
+- Import the module in the [init.go](https://github.com/netdata/go.d.plugin/blob/master/modules/init.go)
```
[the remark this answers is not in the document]
georgeok: done

modules/mongodb/README.md
```diff
@@ -0,0 +1,342 @@
...
+
+It collects information and statistics about the server executing the following commands:
+
+- [`serverStatus`](https://docs.mongodb.com/manual/reference/command/serverStatus/#mongodb-dbcommand-dbcmd.serverStatus)
+
+## Charts
+
+### Default chart
+
+Works with local and Atlas database servers
```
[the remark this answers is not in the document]
georgeok: done

modules/mongodb/collect.go
```diff
@@ -0,0 +1,104 @@
...
+	"context"
+	"strings"
+	"time"
+
+	"github.com/netdata/go.d.plugin/agent/module"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+func addIfExists(serverStatus map[string]interface{}, id string, ms map[string]int64) {
+	mMap := serverStatus
```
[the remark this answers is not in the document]
georgeok: done

modules/mongodb/collect.go
```diff
@@ -0,0 +1,104 @@
...
+	m.metricExists(status, "wiredTiger.cache", &chartWiredTigerCache)
+	m.metricExists(status, "wiredTiger.capacity", &chartWiredTigerCapacity)
+	m.metricExists(status, "wiredTiger.connection", &chartWiredTigerConnection)
+	m.metricExists(status, "wiredTiger.cursor", &chartWiredTigerCursor)
+	m.metricExists(status, "wiredTiger.lock", &chartWiredTigerLock)
+	m.metricExists(status, "wiredTiger.lock", &chartWiredTigerLockDuration)
+	m.metricExists(status, "wiredTiger.log", &chartWiredTigerLogOps)
+	m.metricExists(status, "wiredTiger.log", &chartWiredTigerLogBytes)
+	m.metricExists(status, "wiredTiger.transaction", &chartWiredTigerTransactions)
+}
```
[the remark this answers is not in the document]
georgeok: The problem with adding all the charts once is that mongo doesn't report any metrics until there are some actual data to show and then subsequently it always reports the data. If we add all the wiredTiger charts, we have the risk of having some charts with empty data for a while or forever (depending on the database version in the future)

modules/mongodb/collect_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
georgeok: done

modules/mongodb/config.go
georgeok: I've moved the code to `mongodb`

modules/mongodb/testdata/v5.0.0/serverStatus.go
georgeok: done

modules/mongodb/collect_test.go
georgeok: done

modules/mongodb/mongodb.go
georgeok: done

modules/mongodb/mongodb.go
georgeok: done

modules/mongodb/collect.go
```diff
@@ -0,0 +1,104 @@
...
+			serverStatus = val.(map[string]interface{})
+		default:
+			return
+		}
+	}
+	if !m.charts.Has(chart.ID) {
+		if err := m.charts.Add(chart.Copy()); err != nil {
+			m.Warning(err)
+		}
+		return
```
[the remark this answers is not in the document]
georgeok: I have added a map to monitor the additions

modules/mongodb/collect.go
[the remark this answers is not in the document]
ilyam8: Ok, didn't know that. Thanks!

modules/mongodb/mongodb.go
georgeok: I am not sure about this. Ping will actual send a ping command to the database and see if it's live or not. We do that anyway in the `Check` method where we call `Collect` and this will fail if the database is don't or configuration is wrong. I was actually thinking do a ping in the check method instead of collect because it's more lightweight. What do you think?

modules/mongodb/README.md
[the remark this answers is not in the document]
georgeok: I am not 100% sure about the ping. It does send a command to the database and adds a round trip. I was actually thinking replacing the `collect` with `ping` in the check method. If we do just the connect as we do it now, check will fail anyway because it calls collect if the database is down or the connection uri is wrong. What do you think?

modules/mongodb/mongodb.go
georgeok: I am not 100% sure about the ping. It does send a command to the database and adds a round trip. I was actually thinking replacing the `collect` with `ping` in the check method. If we do just the connect as we do it now, check will fail anyway because it calls collect if the database is down or the connection uri is wrong. What do you think?

ilyam8: That is a good idea. One possible issue I can think of - ping (check) success, but collect fails, because of permissions.

ilyam8: We consider a job active the job's Check returns `true` and keep retrying (call Collect) it indefinitely (but there is a retry backoff algorithm i believe). So in that case we get "check ok" in the logs, but no charts on the dashboard.

georgeok: Ok, it makes sense to leave the collect as it is then. Regarding the ping I think we don't need it.

ilyam8: I agree! Not sure why I brought up this question 😅

modules/mongodb/mongodb.go
georgeok: done

modules/mongodb/mongodb_test.go
georgeok: done

modules/mongodb/README.md
```diff
@@ -0,0 +1,342 @@
...
+- rollovers in `asserts/s`
+
+#### Current Transactions
+
+- current active in `transactions`
+- current inactive in `transactions`
+- current open in `transactions`
+- current prepared in `transactions`
+
+### Optional charts:
```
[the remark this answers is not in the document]
georgeok: done

ilyam8:
@georgeok hey, reviewing charts. I have few general suggestions:
 - let's don't use defaults (module.Line for chart.Type, and module.Absolute for dimensions.Algo).
 - let's group charts into different `var ( ... )` block by families (chart.Fam).

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+	Title: "Current Transactions",
+	Units: "transactions",
+	Fam:   "transactionsCurrent",
+	Ctx:   "mongodb.transactionsCurrent",
+	Dims: module.Dims{
+		{ID: "transactions.currentActive", Name: "current active"},
+		{ID: "transactions.currentInactive", Name: "current inactive"},
+		{ID: "transactions.currentOpen", Name: "current open"},
+		{ID: "transactions.currentPrepared", Name: "current prepared"},
+	},
```
ilyam8, line 169:
```suggestion
var chartTransactionsCurrent = module.Chart{
	ID:    "transactionsCurrent",
	Title: "Current Transactions",
	Units: "transactions",
	Fam:   "transactionsCurrent",
	Ctx:   "mongodb.current_transactions",
	Dims: module.Dims{
		{ID: "transactions.currentActive", Name: "active"},
		{ID: "transactions.currentInactive", Name: "inactive"},
		{ID: "transactions.currentOpen", Name: "open"},
		{ID: "transactions.currentPrepared", Name: "prepared"},
	},
```

- "snake_case" Ctx
- no need in "current " in the names, the chart title is "Current Transactions"

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+
+var chartGlobalLockActiveClients = module.Chart{
+	ID:    "globalLockActiveClients",
+	Title: "Active Clients",
+	Units: "clients",
+	Fam:   "clients",
+	Ctx:   "mongodb.currentQueue",
+	Dims: module.Dims{
+		{ID: "globalLock.activeClients.readers", Name: "readers"},
+		{ID: "globalLock.activeClients.writers", Name: "writers"},
```
ilyam8, line 180:
```suggestion
var chartGlobalLockActiveClients = module.Chart{
	ID:    "globalLockActiveClients",
	Title: "Active Clients",
	Units: "clients",
	Fam:   "clients",
	Ctx:   "mongodb.active_clinents",
	Dims: module.Dims{
		{ID: "globalLock.activeClients.readers", Name: "readers"},
		{ID: "globalLock.activeClients.writers", Name: "writers"},
```

The metric is "active_clients" (performing read/write operations)

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+	},
+}
+
+var chartCollections = module.Chart{
+	ID:    "catalogStats",
+	Title: "Catalog Stats",
+	Units: "objects",
+	Fam:   "catalogStats",
+	Ctx:   "mongodb.catalogStats",
```
ilyam8, line 189: What is "catalog"? Didn't find it in [the glossary](https://docs.mongodb.com/manual/reference/glossary/) or [serverStatus](https://docs.mongodb.com/manual/reference/command/serverStatus).

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+	chartTcmallocGeneric = module.Chart{
+		ID:    "tcmallocGeneric",
+		Title: "Tcmalloc generic metrics",
+		Units: "MiB",
+		Fam:   "tcmalloc",
+		Ctx:   "mongodb.tcmallocGeneric",
+		Dims: module.Dims{
+			{ID: "tcmalloc.generic.current_allocated_bytes", Name: "current_allocated_bytes", Div: 1 << 20},
+			{ID: "tcmalloc.generic.heap_size", Name: "heap_size", Div: 1 << 20},
+		},
```
ilyam8, line 211:
```suggestion
	chartTcmallocGeneric = module.Chart{
		ID:    "tcmallocGeneric",
		Title: "Tcmalloc generic metrics",
		Units: "bytes",
		Fam:   "memory",
		Ctx:   "mongodb.tcmalloc_generic",
		Dims: module.Dims{
			{ID: "tcmalloc.generic.current_allocated_bytes", Name: "current_allocated_bytes"},
			{ID: "tcmalloc.generic.heap_size", Name: "heap_size"},
		},
```

- didn't find these metrics in [serverStatus](https://docs.mongodb.com/manual/reference/command/serverStatus/)
- I guess it makes sense to move it to "memory"
- it is in bytes, correct? If so there is no need to convert to something else
- "snake_case" Ctx

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+		Dims: module.Dims{
+			{ID: "tcmalloc.tcmalloc.pageheap_free_bytes", Name: "Pageheap free", Div: 1024},
+			{ID: "tcmalloc.tcmalloc.pageheap_unmapped_bytes", Name: "Pageheap unmapped ", Div: 1024},
+			{ID: "tcmalloc.tcmalloc.current_total_thread_cache_bytes", Name: "Total threaded cache", Div: 1024},
+			{ID: "tcmalloc.tcmalloc.total_free_bytes", Name: "Free", Div: 1024},
+			{ID: "tcmalloc.tcmalloc.pageheap_committed_bytes", Name: "Pageheap committed", Div: 1024},
+			{ID: "tcmalloc.tcmalloc.pageheap_total_commit_bytes", Name: "Pageheap total commit", Div: 1024},
+			{ID: "tcmalloc.tcmalloc.pageheap_total_decommit_bytes", Name: "Pageheap decommit", Div: 1024},
+			{ID: "tcmalloc.tcmalloc.pageheap_total_reserve_bytes", Name: "Pageheap reserve", Div: 1024},
+		},
```
ilyam8, line 228:
```suggestion
	chartTcmalloc = module.Chart{
		ID:    "tcmalloc",
		Title: "Tcmalloc",
		Units: "bytes",
		Fam:   "memory",
		Ctx:   "mongodb.tcmalloc",
		Dims: module.Dims{
			{ID: "tcmalloc.tcmalloc.pageheap_free_bytes", Name: "Pageheap free"},
			{ID: "tcmalloc.tcmalloc.pageheap_unmapped_bytes", Name: "Pageheap unmapped "},
			{ID: "tcmalloc.tcmalloc.current_total_thread_cache_bytes", Name: "Total threaded cache"},
			{ID: "tcmalloc.tcmalloc.total_free_bytes", Name: "Free"},
			{ID: "tcmalloc.tcmalloc.pageheap_committed_bytes", Name: "Pageheap committed"},
			{ID: "tcmalloc.tcmalloc.pageheap_total_commit_bytes", Name: "Pageheap total commit"},
			{ID: "tcmalloc.tcmalloc.pageheap_total_decommit_bytes", Name: "Pageheap decommit"},
			{ID: "tcmalloc.tcmalloc.pageheap_total_reserve_bytes", Name: "Pageheap reserve"},
		},
```

- didn't find these metrics in [serverStatus](https://docs.mongodb.com/manual/reference/command/serverStatus/)
- I guess it makes sense to move it to "memory"
- it is in bytes, correct? If so there is no need to convert to something else

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+	Title: "Current Queue Clients",
+	Units: "clients",
+	Fam:   "clients",
+	Ctx:   "mongodb.currentQueue",
+	Type:  module.Stacked,
+	Dims: module.Dims{
+		{ID: "globalLock.currentQueue.readers", Name: "readers"},
+		{ID: "globalLock.currentQueue.writers", Name: "writers"},
+	},
+}
```
ilyam8, line 243:
```suggestion
var chartGlobalLockCurrentQueue = module.Chart{
	ID:    "globalLockCurrentQueue",
	Title: "Queued operations because of a lock",
	Units: "operations",
	Fam:   "clients",
	Ctx:   "mongodb.queued_operations",
	Type:  module.Stacked,
	Dims: module.Dims{
		{ID: "globalLock.currentQueue.readers", Name: "readers"},
		{ID: "globalLock.currentQueue.writers", Name: "writers"},
	},
}
```

- I see [globalLock.currentQueue](https://docs.mongodb.com/manual/reference/command/serverStatus/#mongodb-serverstatus-serverstatus.globalLock.currentQueue) is the number of queued operations, not clients (units)
- also, let's move it to "clients" `var (...)` section

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+		{ID: "metrics.commands.eval.total", Name: "Eval", Algo: module.Incremental},
+		{ID: "metrics.commands.eval.failed", Name: "Eval Failed", Algo: module.Incremental},
+		{ID: "metrics.commands.delete.total", Name: "Delete", Algo: module.Incremental},
+		{ID: "metrics.commands.delete.failed", Name: "Delete Failed", Algo: module.Incremental},
+		{ID: "metrics.commands.count.failed", Name: "Count Failed", Algo: module.Incremental},
+		{ID: "metrics.commands.createIndexes", Name: "Create Indexes", Algo: module.Incremental},
+		{ID: "metrics.commands.findAndModify", Name: "Find And Modify", Algo: module.Incremental},
+		{ID: "metrics.commands.insert.failed", Name: "Insert Fail", Algo: module.Incremental},
+	},
+}
```
ilyam8, line 261: These are [database commands](https://docs.mongodb.com/manual/reference/command/), and there are a lot of commands. Why do we need these and not others?

Also, I think it'd better have 2 charts: executed and failed.

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+	Fam:   "locks",
+	Ctx:   "mongodb.locks",
+	Dims: module.Dims{
+		{ID: "locks.Global.acquireCount.r", Name: "Global Read Locks", Algo: module.Incremental},
+		{ID: "locks.Global.acquireCount.w", Name: "Global Write Locks", Algo: module.Incremental},
+		{ID: "locks.Database.acquireCount.r", Name: "Database Read Locks", Algo: module.Incremental},
+		{ID: "locks.Database.acquireCount.w", Name: "Database Write Locks", Algo: module.Incremental},
+		{ID: "locks.Collection.acquireCount.r", Name: "Collection Read Locks", Algo: module.Incremental},
+		{ID: "locks.Collection.acquireCount.w", Name: "Collection Write Locks", Algo: module.Incremental},
+	},
```
ilyam8, line 276:
```suggestion
var chartGlobalLocks = module.Chart{
	ID:    "locks",
	Title: "Acquired locks",
	Units: "locks/s",
	Fam:   "locks",
	Ctx:   "mongodb.locks",
	Dims: module.Dims{
		{ID: "locks.Global.acquireCount.r", Name: "global read", Algo: module.Incremental},
		{ID: "locks.Global.acquireCount.w", Name: "global write", Algo: module.Incremental},
		{ID: "locks.Database.acquireCount.r", Name: "database read", Algo: module.Incremental},
		{ID: "locks.Database.acquireCount.w", Name: "database write", Algo: module.Incremental},
		{ID: "locks.Collection.acquireCount.r", Name: "collection read", Algo: module.Incremental},
		{ID: "locks.Collection.acquireCount.w", Name: "collection write", Algo: module.Incremental},
	},
```

I think we can remove " Locks", it is clear from the context (the chart title).

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+
+var chartFlowControl = module.Chart{
+	ID:    "flowControl",
+	Title: "Flow Control Stats",
+	Units: "msec",
+	Fam:   "flowControl",
+	Ctx:   "mongodb.flowControl",
+	Dims: module.Dims{
+		{ID: "flowControl.timeAcquiringMicros", Name: "timeAcquiring", Algo: module.Incremental, Div: 1000},
+		{ID: "flowControl.isLaggedTimeMicros", Name: "isLaggedTime", Algo: module.Incremental, Div: 1000},
```
ilyam8, line 287:
```suggestion
var chartFlowControl = module.Chart{
	ID:    "flowControl",
	Title: "Flow Control Stats",
	Units: "milliseconds",
	Fam:   "flowControl",
	Ctx:   "mongodb.flow_control_timings",
	Dims: module.Dims{
		{ID: "flowControl.timeAcquiringMicros", Name: "timeAcquiring", Algo: module.Incremental, Div: 1000},
		{ID: "flowControl.isLaggedTimeMicros", Name: "isLaggedTime", Algo: module.Incremental, Div: 1000},
```

- let's prefer "milliseconds" over "msec"
- I don't really like "flow_control_timings", but that is less generic than just "flow_control"

ilyam8:
@georgeok 

1. I think we should avoid testing implementation details ([private functions](https://github.com/georgeok/go.d.plugin/blob/6ec6d36b982a151021e75a41c06abae2c217f6fb/modules/mongodb/collect_test.go#L15-L25)), we need to test public methods - Collect. Implementation can be changed, testing using private functions means rewriting test. Not the best example for your case, but should explain the gist - see [HAProxy Collect tests](https://github.com/netdata/go.d.plugin/blob/3e6dd58663d8d5a4803fc2cdb9f433c411d91f52/modules/haproxy/haproxy_test.go#L121-L160).

2. This makes things more difficult, but I think we need to normalize metrics (no camelCase, lowercase, replace dots with underscores). Other monitoring solutions do it.

One way to do it is using [`stm`](https://github.com/netdata/go.d.plugin/tree/master/pkg/stm#stm) package (it requires creating that big serverStatus struct, but it is very easy to convert it into map[string]int64 with good metric names).

modules/mongodb/collect.go
```diff
@@ -0,0 +1,107 @@
...
+			return
+		}
+		switch typeVal := val.(type) {
+		case map[string]interface{}:
+			serverStatus = typeVal
+		default:
+			return
+		}
+	}
+	if enabled, ok := m.optionalChartsEnabled[chart.ID]; !ok || !enabled {
```
ilyam8, line 64:
```suggestion
	if !m.optionalChartsEnabled[chart.ID] {
```

We never change the value to `false`, I think we can simplify this part

modules/mongodb/charts.go
georgeok: correct, they are in bytes

modules/mongodb/charts.go
georgeok: I think that's a v5 and it's not documented. References can be found in their Jira https://jira.mongodb.org/browse/SERVER-55784?jql=text%20~%20%22catalogStats%22

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+	//chartWiredTigerBlockManager.Copy(),
+	//chartWiredTigerCache.Copy(),
+	//chartWiredTigerCapacity.Copy(),
+	//chartWiredTigerConnection.Copy(),
+	//chartWiredTigerCursor.Copy(),
+	//chartWiredTigerLock.Copy(),
+	//chartWiredTigerLockDuration.Copy(),
+	//chartWiredTigerLogOps.Copy(),
+	//chartWiredTigerLogBytes.Copy(),
+	//chartWiredTigerTransactions.Copy(),
```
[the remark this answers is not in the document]
georgeok: done

modules/mongodb/charts.go
```diff
@@ -0,0 +1,490 @@
...
+	Dims: module.Dims{
+		{ID: "connections.current", Name: "current"},
+		{ID: "connections.active", Name: "active"},
+		{ID: "connections.threaded", Name: "threaded"},
+		{ID: "connections.exhaustIsMaster", Name: "exhaustIsMaster"},
+		{ID: "connections.exhaustHello", Name: "exhaustHello"},
+		{ID: "connections.awaitingTopologyChanges", Name: "awaiting topology changes"},
+		{ID: "connections.available", Name: "available"},
+	},
+}
```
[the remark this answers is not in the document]
georgeok: I guess it makes sense. Given the fact it's my first contribution, I'll listen to the experience 😉

modules/mongodb/charts.go
georgeok: I think that's one the non-documented features too. I had to trace it in the source code here https://github.com/mongodb/mongo/blob/master/src/mongo/util/tcmalloc_server_status_section.cpp

modules/mongodb/collect.go
georgeok: correct, I forgot that Go return the default value in such cases

modules/mongodb/charts.go
ilyam8: Thanks! How did you find it? 😄 

Ok, "Motivation" section from the Jira ticker

> In order to make more informed product decisions and better support customers feature requests we need to have better insight into our customer usage and behavior. By adding metrics to track usage and behavior ASAP we will be able to not only measure the success of features and inits, but better understand how to develop and prioritize new features related to collection types. 

Do we need these stats? It looks like they (product) need it

> better understand how to develop and prioritize new features related to collection types.

modules/mongodb/charts.go
georgeok: We have now removed the command to save on the request size.
Now we do `command := bson.D{{Key: "serverStatus", Value: 1}, {Key: "metrics", Value: 0}, {Key: "repl", Value: 0}}` to half the size of the response

georgeok:
I've done quite a refactor:

- Normalised IDs for charts and Dims
- Remove tests for private functions
- Added a big struct with the response schema
- Using the `stm` package for the ID mapping
- Created a database abstraction layer in order to mock the DB in the test

PR #603, opened 2021-08-17
fix some mongodb serverStatus charts

ilyam8:
I noticed that the following charts have dimensions with `-` values:
 - "locks"
 - "wiredtiger_lock"
 - "wiredtiger_log_ops"

This PR fixes the problem.

cc @georgeok


georgeok:
Nice catch

Issue #594, opened 2021-07-15
add MongoDB collector


ilyam8:
We need to rewrite [python mongodb collector](https://github.com/netdata/netdata/tree/master/collectors/python.d.plugin/mongodb#mongodb-monitoring-with-netdata) in Golang.

There are several ways to do it:
 1. _mirror_ python module
 2. rewrite from scratch

### mirror python module

This means doing the same queries, collecting the same metrics, creating the same charts. The problem with it - python version is outdated.

### rewrite from scratch (preferred)

This means to do some research first:
 - read about [MongoDB monitoring](https://docs.mongodb.com/manual/administration/monitoring/) (single setup, replication, sharding, etc.).
 - check existing open-source MongoDB collectors:
   - [percona/mongodb_exporter](https://github.com/percona/mongodb_exporter)
   - [telegraf/mongodb](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mongodb#mongodb-input-plugin)
   - etc.
 - do a proposal (what queries we need, sample output, how to group metrics into charts, etc.)
 - implement according to the proposal (split the work into several PRs if needed (likely yes))

---

georgeok:
Having played with mongo for a while now I propose to get 4 pillars of metrics and 4 PR accordingly:

1. serverStatus which return a general view of the database.
2. dbStats which reflect to the storage use and and data volumes. This one has a lot of parameters and appropriate grouping must be evaluated carefully to see where we get the most value without filling the dashboard with charts.
3. replSetGetStatus to the see the replica set status
4.  sharded cluster status


Initially until we have the basic of the plugin working with multiple database and authentication, I suggest to start with `serverStatus` and group the metrics into 2 charts, one with the storage parameters(disk, ram usage, etc.) in bytes and one with the storage counters (collections, indexes, view, etc.) as the first PR

ilyam8:
Looks good @georgeok, I agree with both:
 - a PR per area (server stats, db stats, etc.).
 - start with server stats.

Not sure about 

> into 2 charts

I will check [the PR](https://github.com/netdata/go.d.plugin/pull/598)

ilyam8:
Related https://github.com/netdata/netdata/issues/5897

georgeok:
@ilyam8 
For the second part of this issue, `dbstats` I suggest that we collect the following metrics:

Aggregated for all databases
- absolute number of collections, views, indexes (1 chart)
-  absolute number of documents  (1 chart)
[...]

PR #604, opened 2021-08-18
ReplicaSetStatus charts

georgeok:
Part of #594

Metrics for replica sets including:

- Replication Lag
- Heartbeat latency
- Ping time

### DIFF
--- a/modules/mongodb/charts.go
+++ b/modules/mongodb/charts.go
@@ -576,37 +576,25 @@ func (m *Mongo) removeDatabasesFromDBStatsCharts(newDatabases []string) {
 func (m *Mongo) removeReplicaSetMembers(newMembers []string) {
 	diff := sliceDiff(m.replSetMembers, newMembers)
 	for _, name := range diff {
-		dimID := replicationHeartbeatLatencyDimPrefix + name
-		chart := m.charts.Get(replicationHeartbeatLatency)
-		if chart != nil {
-			if err := chart.MarkDimRemove(dimID, true); err != nil {
-				m.Warningf("failed to remove dimension: %v", err)
+		for _, v := range []struct{ chartID, dimPrefix string }{
+			{replicationLag, replicationLagDimPrefix},
+			{replicationHeartbeatLatency, replicationHeartbeatLatencyDimPrefix},
+			{replicationNodePing, replicationNodePingDimPrefix},
+		} {
+			id := v.dimPrefix + name
+			if !m.replSetDimsEnabled[id] {
+				continue
 			}
-			delete(m.replSetDimsEnabled, dimID)
-		} else {
-			m.Warningf("failed to remove dimension:%s. job doesn't have chart: %s", dimID, replicationHeartbeatLatency)
-		}
+			delete(m.replSetDimsEnabled, id)
 
-		dimID = replicationLagDimPrefix + name
-		chart = m.charts.Get(replicationLag)
-		if chart != nil {
-			if err := chart.MarkDimRemove(dimID, true); err != nil {
-				m.Warningf("failed to remove dimension: %v", err)
+			chart := m.charts.Get(v.chartID)
+			if chart != nil {
+				if err := chart.MarkDimRemove(id, true); err != nil {
+					m.Warningf("failed to remove dimension: %v", err)
+				}
+			} else {
+				m.Warningf("failed to remove dimension:%s. job doesn't have chart: %s", id, v.chartID)
 			}
-			delete(m.replSetDimsEnabled, dimID)
-		} else {
-			m.Warningf("failed to remove dimension:%s. job doesn't have chart: %s", dimID, replicationLag)
-		}
-
-		dimID = replicationNodePingDimPrefix + name
-		chart = m.charts.Get(replicationNodePing)
-		if chart != nil {
-			if err := chart.MarkDimRemove(dimID, true); err != nil {
-				m.Warningf("failed to remove dimension: %v", err)
-			}
-			delete(m.replSetDimsEnabled, dimID)
-		} else {
-			m.Warningf("failed to remove dimension:%s. job doesn't have chart: %s", dimID, replicationNodePing)
 		}
 	}
 }
--- a/modules/mongodb/charts.go
+++ b/modules/mongodb/charts.go
@@ -577,21 +577,33 @@ func (m *Mongo) removeReplicaSetMembers(newMembers []string) {
 	diff := sliceDiff(m.replSetMembers, newMembers)
 	for _, name := range diff {
 		dimID := replicationHeartbeatLatencyDimPrefix + name
-		err := m.charts.Get(replicationHeartbeatLatency).MarkDimRemove(dimID, true)
-		if err != nil {
-			m.Warningf("failed to remove dimension: %v", err)
+		chart := m.charts.Get(replicationHeartbeatLatency)
+		if chart != nil {
+			if err := chart.MarkDimRemove(dimID, true); err != nil {
+				m.Warningf("failed to remove dimension: %v", err)
+			}
+		} else {
+			m.Warningf("failed to remove dimension:%s. job doesn't have chart: %s", dimID, replicationHeartbeatLatency)
 		}
 
 		dimID = replicationLagDimPrefix + name
-		err = m.charts.Get(replicationLag).MarkDimRemove(dimID, true)
-		if err != nil {
-			m.Warningf("failed to remove dimension: %v", err)
+		chart = m.charts.Get(replicationLag)
+		if chart != nil {
+			if err := chart.MarkDimRemove(dimID, true); err != nil {
+				m.Warningf("failed to remove dimension: %v", err)
+			}
+		} else {
+			m.Warningf("failed to remove dimension:%s. job doesn't have chart: %s", dimID, replicationLag)
 		}
 
 		dimID = replicationNodePingDimPrefix + name
-		err = m.charts.Get(replicationNodePing).MarkDimRemove(dimID, true)
-		if err != nil {
-			m.Warningf("failed to remove dimension: %v", err)
+		chart = m.charts.Get(replicationNodePing)
+		if chart != nil {
+			if err := chart.MarkDimRemove(dimID, true); err != nil {
+				m.Warningf("failed to remove dimension: %v", err)
+			}
+		} else {
+			m.Warningf("failed to remove dimension:%s. job doesn't have chart: %s", dimID, replicationNodePing)
 		}
 	}
 }
--- a/modules/mongodb/charts.go
+++ b/modules/mongodb/charts.go
@@ -27,6 +27,7 @@ var dbStatsCharts = module.Charts{
 	chartDBStatsSize,
 }
 
+// replCharts on used on replicate sets
 var replCharts = module.Charts{
 	chartReplLag,
 	chartReplHeartbeatLatency,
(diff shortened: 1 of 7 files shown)


modules/mongodb/charts.go
```diff
@@ -529,6 +571,31 @@ func (m *Mongo) removeDatabasesFromDBStatsCharts(newDatabases []string) {
...
 
+// removeReplicaSetMember removes dimensions for not existing
+// replicate set members
+func (m *Mongo) removeReplicaSetMembers(newMembers []string) {
+	diff := sliceDiff(m.replSetMembers, newMembers)
+	for _, name := range diff {
+		dimID := replicationHeartbeatLatencyDimPrefix + name
+		err := m.charts.Get(replicationHeartbeatLatency).MarkDimRemove(dimID, true)
+		if err != nil {
+			m.Warningf("failed to remove dimension %s from replication_heartbeat_latency chart: %v", dimID, err)
```
ilyam8, line 582:
```suggestion
			m.Warningf("failed to remove dimension: %v", err)
```

Can be simplified, because the error contains the chart and dimension id's

https://github.com/netdata/go.d.plugin/blob/345a6a148141a72f48c747c40f4ab5c08a881791/agent/module/charts.go#L250-L253

modules/mongodb/charts.go
```diff
@@ -529,6 +571,31 @@ func (m *Mongo) removeDatabasesFromDBStatsCharts(newDatabases []string) {
...
+		}
+
+		dimID = replicationLagDimPrefix + name
+		err = m.charts.Get(replicationLag).MarkDimRemove(dimID, true)
+		if err != nil {
+			m.Warningf("failed to remove dimension %s from replication_lag chart: %v", dimID, err)
+		}
```
ilyam8, line 589:
```suggestion
			m.Warningf("failed to remove dimension: %v", err)
		}
```

same

modules/mongodb/charts.go
```diff
@@ -529,6 +571,31 @@ func (m *Mongo) removeDatabasesFromDBStatsCharts(newDatabases []string) {
...
+
+		dimID = replicationNodePingDimPrefix + name
+		err = m.charts.Get(replicationNodePing).MarkDimRemove(dimID, true)
+		if err != nil {
+			m.Warningf("failed to remove dimension %s from replication_node_ping chart: %v", dimID, err)
```
ilyam8, line 594:
```suggestion
			m.Warningf("failed to remove dimension: %v", err)
```

same

modules/mongodb/charts.go
```diff
@@ -529,6 +571,31 @@ func (m *Mongo) removeDatabasesFromDBStatsCharts(newDatabases []string) {
 	}
 }
 
+// removeReplicaSetMember removes dimensions for not existing
+// replicate set members
+func (m *Mongo) removeReplicaSetMembers(newMembers []string) {
+	diff := sliceDiff(m.replSetMembers, newMembers)
+	for _, name := range diff {
+		dimID := replicationHeartbeatLatencyDimPrefix + name
+		err := m.charts.Get(replicationHeartbeatLatency).MarkDimRemove(dimID, true)
```
ilyam8, line 580: Is it possible that "m.charts.Get(replicationHeartbeatLatency)" returns nil?

modules/mongodb/mongodb.go
```diff
@@ -96,6 +103,24 @@ func (m *Mongo) Collect() map[string]int64 {
...
+		// we add once the charts during runtime
+		m.addReplChartsOnce.Do(func() {
+			replCharts := replCharts.Copy()
+			for _, chart := range *replCharts {
+				err := m.charts.Add(chart)
+				if err != nil {
+					m.Errorf("failed to add replica set chart: %s", chart.ID)
+				}
+			}
+		})
```
ilyam8, line 117:
```suggestion
		m.addReplChartsOnce.Do(func() {
			if err := m.charts.Add(*replCharts.Copy()...); err != nil {
				m.Errorf("failed to add replica set chart: %v", err)
			}
		})
```

Why don't add all the charts at once?

modules/mongodb/collect.go
```diff
@@ -109,3 +109,69 @@ func (m *Mongo) collectDbStats(ms map[string]int64) error {
...
+
+	// replica nodes may be removed
+	// we should collect metrics for these anymore
+	m.removeReplicaSetMembers(currentMembers)
+	m.replSetMembers = currentMembers
+
+	for _, member := range status.Members {
+		// Heartbeat lag calculation
+		if member.LastHeartbeatReceived != nil {
+			id := replicationHeartbeatLatencyDimPrefix + member.Name
```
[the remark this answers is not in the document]
ilyam8: I mean the following

```go
		if member.LastHeartbeatReceived != nil {
			id := replicationHeartbeatLatencyDimPrefix + member.Name
			ms[id] = status.Date.Sub(*member.LastHeartbeatReceived).Milliseconds()
			if !m.replSetDimsEnabled[id] {
				m.replSetDimsEnabled[id] = true
				err := m.charts.Get(replicationHeartbeatLatency).AddDim(&module.Dim{ID: id, Name: member.Name})
				if err != nil {
					m.Warningf("failed to add dim with id: %s, err: %v", id, err)
				}
			}
		}
```

modules/mongodb/charts.go
```diff
@@ -529,6 +571,31 @@ func (m *Mongo) removeDatabasesFromDBStatsCharts(newDatabases []string) {
 	}
 }
 
+// removeReplicaSetMember removes dimensions for not existing
+// replicate set members
+func (m *Mongo) removeReplicaSetMembers(newMembers []string) {
```
ilyam8, line 576: We need to remove the values from "replSetDimsEnabled" in that function too.

ilyam8: Actually, we need to check if an id is in the "replSetDimsEnabled" and delete it only if it is in there (same as we do in the "collectReplSetStatus").

modules/mongodb/charts.go
georgeok: done

modules/mongodb/charts.go
georgeok: done

modules/mongodb/charts.go
georgeok: done

modules/mongodb/mongodb.go
georgeok: good point

modules/mongodb/connector.go
```diff
@@ -56,6 +61,31 @@ func (m *mongoCollector) dbStats(databaseName string) (*dbStats, error) {
...
+	if err != nil {
+		return false
+	}
+	return status.Repl != nil
+}
+
+// replSetGetStatus gets the `replSetGetStatus` from the server
+func (m *mongoCollector) replSetGetStatus() (*replSetStatus, error) {
+	var status *replSetStatus
+	//command := bson.D{{Key: "replSetGetStatus", Value: 1}}
```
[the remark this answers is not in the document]
georgeok: no, I forgot to remove it earlier

modules/mongodb/charts.go
georgeok: yes, that is possible if we fail to add the chart for some reason. A little bit of an edge case because we use constants to add the charts, but I've added nil checks to be 100% sure that the code will never panic and some logging if the chart is nil

modules/mongodb/collect.go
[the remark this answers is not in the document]
georgeok: done

modules/mongodb/charts.go
georgeok: done!

modules/mongodb/charts.go
```diff
@@ -529,6 +571,46 @@ func (m *Mongo) removeDatabasesFromDBStatsCharts(newDatabases []string) {
...
+	diff := sliceDiff(m.replSetMembers, newMembers)
+	for _, name := range diff {
+		dimID := replicationHeartbeatLatencyDimPrefix + name
+		chart := m.charts.Get(replicationHeartbeatLatency)
```
ilyam8, line 580: I think we need to check if "dimID" is in the "m.replSetDimsEnabled[id]" and only then try to get chart/delete the dimension.

I see that [lastHeartbeatRecv](lastHeartbeatRecv) and [pingMs](https://docs.mongodb.com/manual/reference/command/replSetGetStatus/#mongodb-data-replSetGetStatus.members-n-.pingMs) are not available for all replica set members.

Something like

```go
	diff := sliceDiff(m.replSetMembers, newMembers)
	for _, name := range diff {
		for _, v := range []struct{ chartID, dimPrefix string }{
			{chartReplLag.ID, replicationLagDimPrefix},
			{chartReplHeartbeatLatency.ID, replicationHeartbeatLatencyDimPrefix},
			{chartReplPing.ID, replicationNodePingDimPrefix},
		} {
			id := v.dimPrefix + name
			if !m.replSetDimsEnabled[id] {
				continue
			}
            delete(m.replSetDimsEnabled, id)
...
```

ilyam8:
@georgeok that is not really related to this PR, but I suggest doing the following:
 - move Collect() logic to collect() func ([example](https://github.com/netdata/go.d.plugin/blob/345a6a148141a72f48c747c40f4ab5c08a881791/modules/traefik/traefik.go#L93-L104))
 - move collect() to the collect.go
 - create a file per query
   - collect_server_status.go
   - collect_dbstats.go
   - collect_replsetstatus.go

modules/mongodb/charts.go
georgeok: sounds good, I'll change it and also use the constants instead of the chart IDs. For example: `replicationLag` instead of `chartReplLag.ID`

modules/mongodb/collect.go
```diff
@@ -109,3 +109,73 @@ func (m *Mongo) collectDbStats(ms map[string]int64) error {
...
 }
+
+// replSetCollect creates the map[string]int64 for the available dims.
+// nil values will be ignored and not added to the map and thus metrics
+// should not appear on the dashboard.
+// if the querying node does not belong to a replica set
+func (m *Mongo) collectReplSetStatus(ms map[string]int64) error {
+	status, err := m.mongoCollector.replSetGetStatus()
+	if err != nil {
+		return fmt.Errorf("error get server status from mongo: %s", err)
```
ilyam8, line 120:
```suggestion
		return fmt.Errorf("error get status of the replica set from mongo: %s", err)
```

modules/mongodb/collect.go
```diff
@@ -109,3 +109,73 @@ func (m *Mongo) collectDbStats(ms map[string]int64) error {
...
+		// Heartbeat lag calculation
+		if member.LastHeartbeatReceived != nil {
+			id := replicationHeartbeatLatencyDimPrefix + member.Name
+			// add dimension if not exists yet
+			if !m.replSetDimsEnabled[id] {
+				m.replSetDimsEnabled[id] = true
+				chart := m.charts.Get(replicationHeartbeatLatency)
+				if chart != nil {
+					if err := chart.AddDim(&module.Dim{ID: id, Name: member.Name}); err != nil {
+						m.Warningf("failed to add dim with id: %s, err: %v", id, err)
```
ilyam8, line 142:
```suggestion
						m.Warningf("failed to add dim: %v", err)
```

The error contains both the chart and dim ids

modules/mongodb/collect.go
```diff
@@ -109,3 +109,73 @@ func (m *Mongo) collectDbStats(ms map[string]int64) error {
...
+
+		// Replica set time diff between current time and time when last entry from the oplog was applied
+		id := replicationLagDimPrefix + member.Name
+		// add dimension if not exists yet
+		if !m.replSetDimsEnabled[id] {
+			m.replSetDimsEnabled[id] = true
+			chart := m.charts.Get(replicationLag)
+			if chart != nil {
+				if err := chart.AddDim(&module.Dim{ID: id, Name: member.Name}); err != nil {
+					m.Warningf("failed to add dim with id: %s, err: %v", id, err)
```
ilyam8, line 157:
```suggestion
				if err := chart.AddDim(&module.Dim{ID: id, Name: member.Name}); err != nil {
					m.Warningf("failed to add dim: %v", err)
```

The error contains both the chart and dim ids

modules/mongodb/collect.go
```diff
@@ -109,3 +109,73 @@ func (m *Mongo) collectDbStats(ms map[string]int64) error {
...
+		// Ping time
+		if member.PingMs != nil {
+			id := replicationNodePingDimPrefix + member.Name
+			// add dimension if not exists yet
+			if !m.replSetDimsEnabled[id] {
+				m.replSetDimsEnabled[id] = true
+				chart := m.charts.Get(replicationNodePing)
+				if chart != nil {
+					if err := chart.AddDim(&module.Dim{ID: id, Name: member.Name}); err != nil {
+						m.Warningf("failed to add dim with id: %s, err: %v", id, err)
```
ilyam8, line 172:
```suggestion
					if err := chart.AddDim(&module.Dim{ID: id, Name: member.Name}); err != nil {
						m.Warningf("failed to add dim: %v", err)
```

The error contains both the chart and dim ids

modules/mongodb/collect.go
georgeok: done

modules/mongodb/collect.go
georgeok: done

ilyam8 (approved):

modules/mongodb/collect.go
georgeok: done

ilyam8:
Thanks @georgeok Good work!

PR #606, opened 2021-08-24
Mongo collector code refactoring

georgeok:
Following from previous PR [comment](https://github.com/netdata/go.d.plugin/pull/604#issuecomment-904428929)
- move Collect() logic to collect() func
- move collect() to the collect.go
- create a file per query
  - collect_server_status.go
  - collect_dbstats.go
  - collect_replsetstatus.go

There are no changes in functionality

Also fixes the bug with the charts reported in the comments by @ilyam8


modules/mongodb/collect.go
```diff
@@ -1,181 +1,49 @@
...
-
-	if status.GlobalLock != nil {
-		m.metricExists(status.GlobalLock.ActiveClients, &chartGlobalLockActiveClients)
-		m.metricExists(status.GlobalLock.CurrentQueue, &chartGlobalLockCurrentQueue)
-	}
-	if status.Tcmalloc != nil {
-		m.metricExists(status.Tcmalloc.Generic, &chartTcmallocGeneric)
-		m.metricExists(status.Tcmalloc.Tcmalloc, &chartTcmalloc)
+	if err := m.collectDbStats(ms); err != nil {
+		m.Errorf("couldn't collecting dbstats metrics: %s", err)
```
ilyam8, line 16: We don't return "nil, err" here, is that on purpose?

modules/mongodb/collect.go
```diff
@@ -1,181 +1,49 @@
...
+			}
+		})
 
-	// filter matching databases and exclude non-matching
-	var databases []string
-	for _, database := range allDatabases {
-		if m.databasesMatcher.MatchString(database) {
-			databases = append(databases, database)
+		if err := m.collectReplSetStatus(ms); err != nil {
+			m.Errorf("couldn't collecting replSetStatus metrics: %s", err)
```
ilyam8, line 29: We don't return "nil, err" here, is that on purpose?

ilyam8:
@georgeok I found a problem in 

https://github.com/netdata/go.d.plugin/blob/85b21c02ef01d59f7c67cafb64e7120f38bec66f/modules/mongodb/mongodb.go#L134-L149

- we need to add a copy of "serverStatusCharts"
- we can add all the "dbStatsCharts" without for loop

```go
	if err := charts.Add(*m.chartsDbStats...); err != nil {
		return nil, err
	}
```

modules/mongodb/collect.go
georgeok: my idea is to return at least the serverStatus metric if they have been collected successfully before this statement. I am not sure if it's correct or not. Do  you think we should return (nil, error) in such cases?

georgeok:
> @georgeok I found a problem in
> 
> https://github.com/netdata/go.d.plugin/blob/85b21c02ef01d59f7c67cafb64e7120f38bec66f/modules/mongodb/mongodb.go#L134-L149
> 
> * we need to add a copy of "serverStatusCharts"
> * we can add all the "dbStatsCharts" without for loop
> 
> ```go
> 	if err := charts.Add(*m.chartsDbStats...); err != nil {
> 		return nil, err
> 	}
> ```

problem fixed

modules/mongodb/collect.go
ilyam8: > my idea is to return at least the serverStatus 

Yeah, that makes sense. Then we need to return "ms, error" and remove https://github.com/georgeok/go.d.plugin/blob/2c3c75bd5dde92d9c28074f7d9b271e92f7ca59e/modules/mongodb/mongodb.go#L94

georgeok: done

ilyam8 (approved):

ilyam8:
Big thanks for the good work, @georgeok!

PR #607, opened 2021-08-29
mysql+mariadb Process List Query Statistics

georgeok:
Resolves netdata/netdata#7879


modules/mysql/charts.go
```diff
@@ -462,6 +462,34 @@ var charts = Charts{
 			{ID: "open_tables", Name: "tables"},
 		},
 	},
+	{
+		ID:    "process_list_fetch_duration",
+		Title: "Process List Fetch Duration",
+		Units: "milliseconds",
+		Fam:   "process list",
+		Ctx:   "mysql.process_fetch_query_duration",
+		Dims:  Dims{{ID: "process_fetch_query_duration", Name: "Query Duration"}},
```
ilyam8, line 471:
```suggestion
		Ctx:   "mysql.process_list_fetch_query_duration",
		Dims:  Dims{{ID: "process_list_fetch_query_duration", Name: "duration"}},
```

⚠️ don't submit the changes via GitHub

- let's add **_list** (other metrics have it)
- lowercase dimension name. Also, i think just "duration" is ok

modules/mysql/charts.go
```diff
@@ -462,6 +462,34 @@ var charts = Charts{
...
+		ID:    "process_list_query_count",
+		Title: "Queries Count",
+		Units: "queries",
+		Fam:   "process list",
+		Ctx:   "mysql.process_list_query_count",
+		Type:  module.Stacked,
+		Dims: Dims{
+			{ID: "process_list_query_count_system", Name: "System"},
+			{ID: "process_list_query_user", Name: "Users"},
+		},
```
ilyam8, line 483:
```suggestion
		ID:    "process_list_queries_count",
		Title: "Queries Count",
		Units: "queries",
		Fam:   "process list",
		Ctx:   "mysql.process_list_queries_count",
		Type:  module.Stacked,
		Dims: Dims{
			{ID: "process_list_queries_count_system", Name: "system"},
			{ID: "process_list_queries_user", Name: "user"},
		},
```

⚠️ don't submit the changes via GitHub

- "process_list_query_count" = "process_list_queries_count"
- lowercase dimensions names

modules/mysql/charts.go
```diff
@@ -462,6 +462,34 @@ var charts = Charts{
...
+	},
+	{
+		ID:    "process_list_longest_query_duration",
+		Title: "Longest Query Duration",
+		Units: "seconds",
+		Fam:   "process list",
+		Ctx:   "mysql.process_list_longest_query_duration",
+		Dims:  Dims{{ID: "process_list_longest_query_duration", Name: "Longest Query Duration"}},
```
ilyam8, line 491:
```suggestion
		ID:    "process_list_longest_query_duration",
		Title: "Longest Query Duration",
		Units: "milliseconds",
		Fam:   "process list",
		Ctx:   "mysql.process_list_longest_query_duration",
		Dims:  Dims{{ID: "process_list_longest_query_duration", Name: "duration"}},
```

⚠️ don't submit the changes via GitHub

- let's have it in "milliseconds. Netdata automatically converts seconds to "DD:HH:MM:SS" (we don't want to lose milliseconds)
- lowercase the dimension name

modules/mysql/collect.go
```diff
@@ -64,6 +64,10 @@ func (m *MySQL) collect() (map[string]int64, error) {
 		}
 	}
 
+	if err := m.collectProcessListStatistics(collected); err != nil {
+		m.Errorf("error on collecting process list statistics: %v", err)
+	}
```
ilyam8, line 69: I am not 100% sure, but [according to the requeste](https://github.com/netdata/netdata/issues/7879#issuecomment-581942235)r this query requires [PROCESS privileges](https://dev.mysql.com/doc/refman/8.0/en/privileges-provided.html#priv_process).

I think we need to wrap it in another "if/else", like

https://github.com/netdata/go.d.plugin/blob/2181c601562b0d2fa9005756c3d141572f3b0758/modules/mysql/collect.go#L51-L57

But, as the TODO says, ideally we should stop doing the query only on "not enough permissions" error (if that is possible).

modules/mysql/collect_queries_statistics.go
```diff
@@ -0,0 +1,52 @@
...
+
+	var (
+		queryDuration              int64 // queryInfoSchemaProcessList execution time (not including row fetching)
+		longestRunningQuerySeconds int64 // slowest query milliseconds in process list
+		user                       string
+	)
+
+	start := time.Now()
+	rows, err := m.db.Query(queryInfoSchemaProcessList)
+	queryDuration = int64(time.Since(start).Seconds())
```
ilyam8, line 27: let's make it milliseconds

modules/mysql/collect_queries_statistics.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+
+	collected["process_list_query_count_system"] = 0
+	collected["process_list_query_user"] = 0
```
ilyam8, line 34:
```suggestion
	collected["process_list_queries_count_system"] = 0
	collected["process_list_queries_count_user"] = 0
```

⚠️ don't submit the changes via GitHub

Let's rename

modules/mysql/collect_queries_statistics.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		switch user {
+		case "system user", "event_scheduler":
+			collected["process_list_query_count_system"] += 1
+		default:
+			collected["process_list_query_user"] += 1
+		}
+	}
+
+	collected["process_fetch_query_duration"] = queryDuration
```
ilyam8, line 48:
```suggestion
	collected["process_list_fetch_query_duration"] = queryDuration
```

modules/mysql/collect_queries_statistics.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	defer rows.Close()
+
+	collected["process_list_query_count_system"] = 0
+	collected["process_list_query_user"] = 0
+
+	for rows.Next() {
+		if err := rows.Scan(&longestRunningQuerySeconds, &user); err != nil {
+			return err
+		}
```
ilyam8, line 39: We should do it only for the first row

modules/mysql/testdata/mariadb/v10.5.4/process_list.txt
```diff
@@ -0,0 +1,6 @@
+*************************** 1. row ***************************
+TIME: 1
+USER: root
+*************************** 2. row ***************************
+TIME: 10
+USER: system user
```
ilyam8, line 6:
```suggestion
*************************** 1. row ***************************
TIME: 1
USER: root
*************************** 2. row ***************************
TIME: 10
USER: system user
*************************** 3. row ***************************
TIME: 5
USER: system user
```

Let's add another row so we can be sure we really find max TIME and not the last one

modules/mysql/testdata/mariadb/v5.5.46/process_list.txt
```diff
@@ -0,0 +1,6 @@
+*************************** 1. row ***************************
+TIME: 1
+USER: root
+*************************** 2. row ***************************
+TIME: 10
+USER: system user
```
ilyam8, line 6:
```suggestion
*************************** 1. row ***************************
TIME: 1
USER: root
*************************** 2. row ***************************
TIME: 10
USER: system user
*************************** 3. row ***************************
TIME: 5
USER: system user
```
Let's add another row so we can be sure we really find max TIME and not the last one

modules/mysql/testdata/mysql/v8.0.21/process_list.txt
```diff
@@ -0,0 +1,6 @@
+*************************** 1. row ***************************
+TIME: 1
+USER: root
+*************************** 2. row ***************************
+TIME: 10
+USER: system user
```
ilyam8, line 6:
```suggestion
*************************** 1. row ***************************
TIME: 1
USER: root
*************************** 3. row ***************************
TIME: 5
USER: system user
```

Let's add another row so we can be sure we really find max TIME and not the last one

modules/mysql/charts.go
georgeok: `process_list_queries_user should` be `process_list_queries_count_user` following the pattern

modules/mysql/charts.go
georgeok: The underlaying query will still report seconds but I'll convert them to milliseconds. In mariaDB we get the milliseconds `TIME_MS` but mysql only reports seconds. If I differentiate the queries then we will probably cause confusion. I suggest that we stick with the seconds `TIME` column in the query

modules/mysql/charts.go
ilyam8: yeah, you are right, thanks!

modules/mysql/charts.go
ilyam8: > The underlaying query will still report seconds

Ok, I should have paid more attention. `TIME` is "_The time in seconds that the thread has been in its current state._", let's keep it in seconds then.

modules/mysql/collect_queries_statistics.go
georgeok: done, although this will make our unit tests flaky because on some slow machine the mocked duration will be 0 or 1 or a few more milliseconds but I'll figure something out

modules/mysql/collect_queries_statistics.go
georgeok: In our query we order by time `ASC`. The last entry low will also be the maximum time. We could also scan the opposite way and scan the first row only. Not sure what's best.

modules/mysql/testdata/mariadb/v10.5.4/process_list.txt
georgeok: I think this won't work for the same reason as above. Our query return rows ordered by time `ASC`. The last row will have the max time value.

modules/mysql/collect_queries_statistics.go
ilyam8: You are right! We are good if we use `ASC`. Is `ASC` the default one? I see we don't use it explicitly in [the query](https://github.com/georgeok/go.d.plugin/blob/9e7d580dee3f8bfb4398be8d3ae268306f3a8dff/modules/mysql/collect_queries_statistics.go#L11-L13).

modules/mysql/testdata/mariadb/v10.5.4/process_list.txt
ilyam8: Correct, for some reason i was under impressions we do `DESC` 🙈

modules/mysql/collect_queries_statistics.go
georgeok: Yes `ASC` is the default. If I use it explicitly I get a lint warning!

modules/mysql/collect_queries_statistics.go
ilyam8: I think we can just copy the value from `wantCollected`, like i did here https://github.com/netdata/go.d.plugin/blob/2181c601562b0d2fa9005756c3d141572f3b0758/modules/portcheck/portcheck_test.go#L64-L70

modules/mysql/collect_queries_statistics.go
ilyam8: Got it, thanks!

modules/mysql/collect_queries_statistics.go
```diff
@@ -0,0 +1,52 @@
...
+
+	collected["process_list_queries_count_system"] = 0
+	collected["process_list_queries_count_user"] = 0
+
+	for rows.Next() {
+		if err := rows.Scan(&longestRunningQuerySeconds, &user); err != nil {
+			return err
+		}
+		switch user {
+		case "system user", "event_scheduler":
```
ilyam8, line 41: Why those 2? Perhaps to add a link to the doc?

modules/mysql/collect_queries_statistics.go
georgeok: Nice, I did something very similar

modules/mysql/collect_queries_statistics.go
georgeok: I added a link to https://dev.mysql.com/doc/refman/8.0/en/information-schema-processlist-table.html. 
We have two possible values and group them into two groups:

- System
  - system user
  - event_scheduler
- User
  - authenticated
  - unauthenticated

modules/mysql/collect.go
georgeok: I am doing a bit of research and found [this from MariaDB](https://mariadb.com/kb/en/show-processlist/#description):
>SHOW PROCESSLIST shows you which threads are running. You can also get this information from the information_schema.PROCESSLIST table or the mysqladmin processlist command. If you have the PROCESS privilege, you can see all threads. Otherwise, you can see only your own threads (that is, threads associated with the MariaDB account that you are using). If you do not use the FULL keyword, only the first 100 characters of each statement are shown in the Info field.

I tested on the latest MySQL and MariaDB from dockerhub and both give the same result.

- I created an anonymous user
- loged in from the terminal user `mysql -u anonym`
- executed `SELECT TIME,USER FROM INFORMATION_SCHEMA.PROCESSLIST 
WHERE Info IS NOT NULL AND Info NOT LIKE '%PROCESSLIST%' 
ORDER BY TIME`
- Got back an empty result set without errors. 
- executed `SELECT * FROM INFORMATION_SCHEMA.PROCESSLIST`
```
*************************** 1. row ***************************
     ID: 11
   USER: anonym
   HOST: localhost
     DB: NULL
COMMAND: Query
   TIME: 0
  STATE: executing
   INFO: SELECT * FROM INFORMATION_SCHEMA.PROCESSLIST
1 row in set (0.00 sec)
```


**Conclusion:**
It seems like the behaviour is confirmed and the user without the `process` privilege only gets back his own thread and in this was one executing `SELECT * FROM INFORMATION_SCHEMA.PROCESSLIST`. 

If we want stats for all users, I think the only way to resolve this is to have another query at the beginning listing the privileges and checking if process is one of them for the current/netdata user**. Something like:
```
SELECT EXISTS(SELECT PRIVILEGE_TYPE FROM information_schema.USER_PRIVILEGES WHERE GRANTEE='''netdata''@''%''' and PRIVILEGE_TYPE='PROCESS') as HAS_PROCESS_PRIVILEGE;
```
Otherwise we will get the best, for all users if our user has the privilege or just metrics for our user.

ilyam8: > If we want stats for all users, I think the only way to resolve this is to have another query at the beginning listing the privileges and checking if the process is one of them

What privileges do we need to check it? 😄 

---

Actually, that is a good idea. If the user has no needed privileges, there is no sense in doing that query. We need to log an info message and disable the query (not in this PR, I need to think about it).

modules/mysql/README.md
```diff
@@ -80,6 +81,9 @@ It produces the following charts:
 - Connection Errors in `errors/s`
 - Opened Tables in `tables/s`
 - Open Tables in `tables`
+- Process List Fetch Duration in `seconds`
```
ilyam8, line 84:
```suggestion
- Process List Fetch Duration in `milliseconds`
```

modules/mysql/mysql_test.go
```diff
@@ -860,17 +882,31 @@ func TestMySQL_Collect(t *testing.T) {
...
 	}
 
+	// workaround because we measure execution time and that can vary
+	// a few milliseconds depending on the available processing power
+	copyProcessListFetchQueryDuration := func(dst map[string]int64, val int64) {
+		for k := range dst {
+			if strings.HasSuffix(k, "process_list_fetch_query_duration") {
+				dst[k] = val
+			}
+		}
```
ilyam8, line 900: we don't really need to "for" here, but that is minor

ilyam8 (approved):

modules/mysql/README.md
georgeok: sorry, I missed this one

PR #609, opened 2021-09-03
mongodb shard charts and dimensions

georgeok:
Adds mongdb shard metrics.
Part of #594 and resolves netdata/netdata#5897.

### DIFF
--- a/modules/mongodb/collect_shard.go
+++ b/modules/mongodb/collect_shard.go
@@ -1,0 +1,75 @@
+package mongo
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/netdata/go.d.plugin/agent/module"
+)
+
+// collectShard adds the sharding stats for the available dims.
+func (m *Mongo) collectShard(ms map[string]int64) error {
+	// nodes in the shard cluster
+	nodes, err := m.mongoCollector.shardNodes()
+	if err != nil {
+		return err
+	}
+	ms["shard_nodes_count_aware"] = nodes.ShardAware
+	ms["shard_nodes_count_unaware"] = nodes.ShardUnaware
+
+	// databases partitioning
+	databasesPartitioning, err := m.mongoCollector.shardDatabasesPartitioning()
+	if err != nil {
+		return err
+	}
+	ms["shard_databases_partitioned"] = databasesPartitioning.Partitioned
+	ms["shard_databases_unpartitioned"] = databasesPartitioning.UnPartitioned
+
+	// collections partitioning
+	collectionsPartitioning, err := m.mongoCollector.shardCollectionsPartitioning()
+	if err != nil {
+		return err
+	}
+	ms["shard_collections_partitioned"] = collectionsPartitioning.Partitioned
+	ms["shard_collections_unpartitioned"] = collectionsPartitioning.UnPartitioned
+
+	// chunks per shard node
+	chunksPerShard, err := m.mongoCollector.shardChunks()
+	if err != nil {
+		return err
+	}
+	m.updateShartChuckChartDims(chunksPerShard)
+	for shard, count := range chunksPerShard {
+		ms["shard_chucks_per_node_"+shard] = count
+	}
+
+	return nil
+}
+
+func (m *Mongo) updateShartChuckChartDims(chunksPerShard map[string]int64) {
+	chart := m.charts.Get("shard_chucks_per_node")
+	if chart == nil {
+		return
+	}
+	for _, dim := range chart.Dims {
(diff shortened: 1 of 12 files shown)


modules/mongodb/collect_shard.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	ms["shard_collections_partitioned"] = collectionsPartitioning.Partitioned
+	ms["shard_collections_unpartitioned"] = collectionsPartitioning.UnPartitioned
+
+	// chunks per shard node
+	chunksPerShard, err := m.mongoCollector.shardChunks()
+	if err != nil {
+		return err
+	}
+	m.updateShartChuckChartDims(chunksPerShard)
```
ilyam8, line 41:
```suggestion
	m.updateShardChunkChartDims(chunksPerShard)
```

modules/mongodb/collect_shard.go
```diff
@@ -0,0 +1,75 @@
...
+	for shard, count := range chunksPerShard {
+		ms["shard_chucks_per_node_"+shard] = count
+	}
+
+	return nil
+}
+
+func (m *Mongo) updateShartChuckChartDims(chunksPerShard map[string]int64) {
```
ilyam8, line 49:
```suggestion
func (m *Mongo) updateShardChunkChartDims(chunksPerShard map[string]int64) {
```

modules/mongodb/collect_shard.go
```diff
@@ -0,0 +1,75 @@
...
+	chart := m.charts.Get("shard_chucks_per_node")
+	if chart == nil {
+		return
+	}
+	for _, dim := range chart.Dims {
+		if _, ok := chunksPerShard[strings.TrimPrefix(dim.ID, chart.ID+"_")]; !ok {
+			err := chart.MarkDimRemove(dim.ID, true)
+			delete(m.shardNodesDims, dim.ID)
+			if err != nil {
+				m.Error(fmt.Sprintf("updateShartChuckChartDims failed to remove dim %v", err))
```
ilyam8, line 59: prefer Errorf over Error + Sprintf

modules/mongodb/collect_shard.go
```diff
@@ -0,0 +1,75 @@
...
+			}
+		}
+	}
+	for shard := range chunksPerShard {
+		id := chart.ID + "_" + shard
+		if !m.shardNodesDims[id] {
+			err := chart.AddDim(&module.Dim{ID: id, Name: shard, Algo: module.Absolute})
+			m.shardNodesDims[id] = true
+			if err != nil {
```
ilyam8, line 68:
```suggestion
		if !m.shardNodesDims[id] {
			m.shardNodesDims[id] = true
			err := chart.AddDim(&module.Dim{ID: id, Name: shard, Algo: module.Absolute})
			if err != nil {
```

You update ".shardNodesDim" unconditionally, so let's do it right after "if...". Also handling an error right away is better too i think.

modules/mongodb/collect_shard.go
```diff
@@ -0,0 +1,75 @@
...
+				m.Warningf(fmt.Sprintf("updateShartChuckChartDims failed to add dim %v", err))
```
ilyam8, line 69: Sprintf is not needed here, just Warningf.

modules/mongodb/connector.go
```diff
@@ -14,16 +15,28 @@ type connector interface {
...
-// mongoCollector interface that helps abstracting and mocking the database layer.
+// mongoCollector interface that helps to abstract and mock the database layer.
 type mongoCollector struct {
 	Client           *mongo.Client
 	Timeout          time.Duration
 	isReplicaSetFlag *bool
+	isMongosFlag     *bool
+	aggregationFunc  func(
+		client *mongo.Client,
+		ctx context.Context,
```
ilyam8, line 36: Let's make "ctx" the first parameter here (usually that is the pattern)

modules/mongodb/collect_shard.go
georgeok: >Also handling an error right away is better too i think.
I am sure I caught this one. Do you think that we return the error instead of just printing it and return it in the `collectShard()` method?

georgeok: I've add the suggestion above btw

modules/mongodb/collect_shard.go
georgeok: my dyslexia kicked in 😂

modules/mongodb/connector.go
georgeok: yes, I agree. Very good suggestion

PR #619, opened 2021-09-11
coredns 1.7.0 prometheus metrics rename

### LINKED ISSUE #561: `coredns` metric name changes
Since version 1.7.0 of coredns netdata does not collect metrics from coredns for a name change.
The changes are in [this commit](https://github.com/coredns/coredns/commit/19cfa2960cc415e0e38d36fc9d9325bd636c000e#diff-5174d6f6b54aaa256b1c63113e1ae9d4906b1c1b92b9324957468643a5c0e4cd)
New metrics:
from `coredns_panic_count_total` to `coredns_panics_total`
from `coredns_dns_request_count_total` to `coredns_dns_requests_total`
from `coredns_dns_response_rcode_count_total` to `coredns_dns_responses_total`
from `coredns_dns_request_type_count_total` to `coredns_dns_requests_total`

georgeok:
Fixes #561


modules/coredns/collect.go
```diff
@@ -3,30 +3,44 @@ package coredns
 import (
 	"fmt"
 
+	"github.com/hashicorp/go-version"
```
ilyam8, line 6: Can we use

https://github.com/netdata/go.d.plugin/blob/9e912ff0bde663f2319f00bc315140d59c59fc91/go.mod#L11

?

ilyam8:
@georgeok thanks for the fix!

modules/coredns/collect.go
georgeok: sure yes

modules/coredns/collect.go
```diff
@@ -3,30 +3,44 @@ package coredns
...
-	//   coredns_dns_request_count_total{family="1",proto="udp",server="dns://:53",zone="dropped"} 2
-	//   coredns_dns_request_count_total{family="1",proto="udp",server="dns://:53",zone="ya.ru."} 2
-	metricRequestCountTotal       = "coredns_dns_request_count_total"
-	metricRequestTypeCountTotal   = "coredns_dns_request_type_count_total"
-	metricResponseRcodeCountTotal = "coredns_dns_response_rcode_count_total"
+	//   coredns_dns_requests_total{family="1",proto="udp",server="dns://:53",zone="dropped"} 2
+	//   coredns_dns_requests_total{family="1",proto="udp",server="dns://:53",zone="ya.ru."} 2
+	metricRequestCountTotal       string
+	metricRequestTypeCountTotal   string
+	metricResponseRcodeCountTotal string
```
ilyam8, line 43: These are global vars and we change them. Won't work if there are 2 jobs: the first collects from v1.6.9, second from v1.7.0+

georgeok: Yes, that's true. I moved this to a struct and make this struct a variable to the CoreDNS object so that each job will have the correct values

georgeok: @ilyam8 Please let me know if that's ok now...

ilyam8: Looks good 👍 Thanks!

modules/coredns/testdata/version170/some_load.txt
```diff
@@ -0,0 +1,180 @@
...
+coredns_dns_requests_total{family="1",proto="udp",server="dns://:53",zone="coredns.io."} 6
+coredns_dns_requests_total{family="1",proto="udp",server="dns://:53",zone="dropped"} 9
+coredns_dns_requests_total{family="1",proto="udp",server="dns://:53",zone="ya.ru."} 9
+coredns_dns_requests_total{family="1",proto="udp",server="dns://:54",zone="coredns.io."} 13
+coredns_dns_requests_total{family="1",proto="udp",server="dns://:54",zone="dropped"} 12
+coredns_dns_requests_total{family="1",proto="udp",server="dns://:54",zone="ya.ru."} 12
+# HELP coredns_dns_requests_total Counter of DNS requests per type, per zone.
+# TYPE coredns_dns_requests_total counter
+coredns_dns_requests_total{server="",type="A",zone="dropped"} 7
+coredns_dns_requests_total{server="",type="AAAA",zone="dropped"} 7
```
ilyam8, line 19: How did you get this test data? Using `sed`? :smile:

[It should be](https://github.com/coredns/coredns/tree/master/plugin/metrics#description)

- `coredns_dns_requests_total{server, zone, proto, family, type}`

And you have same metric name with different label set:
 - coredns_dns_requests_total{family, proto, server, zone}
 - coredns_dns_requests_total{server, type, zone}

georgeok: I actual tried to cheat and copy the data from v1.6.9 but it didn't work my way. Ok, now last commit should have valid test data.

modules/coredns/collect.go
```diff
@@ -37,6 +51,20 @@ func (cd *CoreDNS) collect() (map[string]int64, error) {
 	}
 
 	mx := newMetrics()
+	if cd.version == nil {
+		cd.version = cd.Version(raw)
+		if cd.version.LTE(version169) {
```
ilyam8, line 56: > if cd.version.LTE(version169)

How does it work if `cd.Version(raw)` returns nil?

georgeok: Atm the moment the collected metric keys will be empty strings which will not provide us any data. We could set a default version but that's also not guaranteed to work if version fails or we could skip the collection of these metrics if we can't find the version. I think the seconds solution is better. What do you think?

ilyam8: What I mean is:
 - "cd.Version" returns nil on error.
 - it means that "if cd.version.LTE(version169)" will panic

modules/coredns/collect.go
```diff
@@ -89,13 +81,34 @@ func (cd *CoreDNS) collect() (map[string]int64, error) {
 	return stm.ToMap(mx), nil
 }
 
-// Summary
+func (cd *CoreDNS) updateVersionDependentMetrics(raw prometheus.Metrics) {
+	if cd.version == nil {
+		if version := cd.Version(raw); version != nil {
```
ilyam8, line 86: Perhaps minor, but still. In case of a version parse error, we can assume that it doesn't make sense to keep trying it - otherwise we will log parse error on every data collection.

modules/coredns/collect.go
```diff
@@ -89,13 +81,34 @@ func (cd *CoreDNS) collect() (map[string]int64, error) {
...
+				cd.metricNames.panicCountTotal = metricPanicCountTotal170orNewer
+				cd.metricNames.requestCountTotal = metricRequestCountTotal170orNewer
+				cd.metricNames.requestTypeCountTotal = metricRequestTypeCountTotal170orNewer
+				cd.metricNames.responseRcodeCountTotal = metricResponseRcodeCountTotal170orNewer
+			}
+		}
+	}
+}
 
 func (cd CoreDNS) Version(raw prometheus.Metrics) *semver.Version {
```
ilyam8, line 103: I suggest making this method private or even make it a function ("parseVersion" or something).

modules/coredns/collect.go
georgeok: oh, I see the issue now. I've pushed a fix along with some small refactoring and a unit test to cover the case where the server doesn't report the version. In this case from now on we will not try to collected the 4 metrics that require the server version.

modules/coredns/collect.go
georgeok: done!

modules/coredns/collect.go
georgeok: I've added some check to avoid retrying. I also had to change some receiver functions from values to pointer because we were passing copies of the `CoreDNS` objects and state changes were lost.
https://github.com/netdata/go.d.plugin/pull/619/files#diff-425bfb25babae1534ae9cac4c982308ab535f71e214b4661abc052a063adfc3dR109
and 
https://github.com/netdata/go.d.plugin/pull/619/files#diff-425bfb25babae1534ae9cac4c982308ab535f71e214b4661abc052a063adfc3dR114

georgeok: Please be careful, I am sure if there if they were value receivers for some specific reason.

modules/coredns/collect.go
```diff
@@ -38,11 +52,19 @@ func (cd *CoreDNS) collect() (map[string]int64, error) {
...
-	cd.collectPanic(mx, raw)
-	cd.collectSummaryRequests(mx, raw)
-	//cd.collectSummaryRequestsDuration(mx, raw)
-	cd.collectSummaryRequestsPerType(mx, raw)
-	cd.collectSummaryResponsesPerRcode(mx, raw)
+	// some metric names are different depending on the version
+	// update them once
+	if !cd.skipVersionCheck {
+		cd.updateVersionDependentMetrics(raw)
+	}
```
ilyam8, line 59: neat: let's change "cd.skipVersionCheck" to true right after "cd.updateVersionDependentMetrics(raw)", not inside it

modules/coredns/collect.go
```diff
@@ -61,14 +83,50 @@ func (cd *CoreDNS) collect() (map[string]int64, error) {
...
+			} else {
+				cd.metricNames.panicCountTotal = metricPanicCountTotal170orNewer
+				cd.metricNames.requestCountTotal = metricRequestCountTotal170orNewer
+				cd.metricNames.requestTypeCountTotal = metricRequestTypeCountTotal170orNewer
+				cd.metricNames.responseRcodeCountTotal = metricResponseRcodeCountTotal170orNewer
+			}
+		}
+	}
+	cd.skipVersionCheck = true
+}
```
ilyam8, line 104: neat: let's use early return, that is always a good idea in my opinion. And I think we don't need to check "if cd.version == nil", that func states that it updates the version, so let's do it (it will be only once anyway because there is "if !cd.skipVersionCheck" check in the collect).

ilyam8:
@georgeok thanks, i have few minor suggestion, apart from that LGTM

georgeok:
All done. Thank you for the valuable feedback

modules/coredns/collect.go
```diff
@@ -38,11 +52,20 @@ func (cd *CoreDNS) collect() (map[string]int64, error) {
...
+		cd.skipVersionCheck = true
+	}
+
+	// we can only get these metrics if we know the server version
+	if cd.version != nil {
+		cd.collectPanic(mx, raw)
+		cd.collectSummaryRequests(mx, raw)
+		cd.collectSummaryRequestsPerType(mx, raw)
+		cd.collectSummaryResponsesPerRcode(mx, raw)
+	}
```
ilyam8, line 68: Wait, if "cd.version" is nil we can't collect anything because "cd.metricNames" fields are empty strings.

modules/coredns/collect.go
```diff
@@ -61,14 +84,49 @@ func (cd *CoreDNS) collect() (map[string]int64, error) {
 	return stm.ToMap(mx), nil
 }
 
-// Summary
+func (cd *CoreDNS) updateVersionDependentMetrics(raw prometheus.Metrics) {
+	version := cd.parseVersion(raw)
+	if version == nil {
+		return
+	}
```
ilyam8, line 91: Should we version = version169 if we can't find a build info metric?

ilyam8: I see **all** "FindByName" functions use "cd.metricNames" and we set the structure fields only if we manage to find/parse the version metric.

I think we always get "coredns_build_info" metric, but if not there are 2 ways

- if we don't try to guess, we should return immidiatly after https://github.com/georgeok/go.d.plugin/blob/5ccfd0b2f30319d5f92076953cac9829528ee54c/modules/coredns/collect.go#L57-L60 if cd.version == nil.
- or assume that version is X and try to process metrics.

I am confused because you added cd.version != check [here](https://github.com/georgeok/go.d.plugin/blob/5ccfd0b2f30319d5f92076953cac9829528ee54c/modules/coredns/collect.go#L62-L68), but not [there](https://github.com/georgeok/go.d.plugin/blob/5ccfd0b2f30319d5f92076953cac9829528ee54c/modules/coredns/collect.go#L70-L81).

ilyam8:
Thanks @georgeok !

PR #624, opened 2021-09-26
socket client

### LINKED ISSUE #622: Pkg/Socket proposal (unify socket client in the agent)
In the agent we have more than one module that are using sockets, and they have their own implementations of TCP clients. 
I propose that we implement a socket client capable of handling TCP, UDP and Unix sockets in order to hide the complexity 
from the individual modules and have a unified way of dealing with socket to make the development of new modules easier.
Something similar to the `pkg/web` package that we have.

At a high level it should be able to send data `[]bytes` over a socket, wait and read the response if any.

Other requirement are:
- ability to handle timeouts
- tls
- ability to perform dns resolution internally of the target is a domain
- handle both IPv4 and IPv6
- have higher level functions to handle string slices
- supported buffered io for better networking performance
- 
An example interface of such a client can be:
```go
type Socket interface {
Connect() error
Disconnect() error
Read(p []byte) (n int, err error)
Write(p []byte) (n int, err error)
}
```
where the caller has to manage the connection and disconnection and call read write.

So that would be 
```go

s := socket{}
s.Connect()
defer s.Disconnect()
s.Write()
s.Read()
```

and/or we can offer some higher level functions
```go
type Socket interface {
	Command([]byte) ([]byte, error)
    CommandStr([]string) ([]string, error)
}
```
where the caller creates a new socket and call command() and the function internally 
figures out everything like DNS resolution, keep-alive, writing, reading, timeouts
and returns some bytes to the caller which then knows what to expect and coverts the
data to appropriate data types. This will also hide `Connect` in case we have `DGRAM` packets.
This is similar to the `fetch()` method in the `zookeeper` module

Maybe we can also support plain packet or ICMP as well, but I don't see much benefit yet for these,
unless we have some modules in mind.

Most of the code can be extracted directly from our `zookeeper` module also be used elsewhere in the future.

Modules that use their own tcp/udp clients:
- zookeeper
- openvpn
- phpfpm
- unbound
and maybe I am missing some more.

Please let me know what your thoughts are

georgeok:
resolves #622

### DIFF
--- a/pkg/socket/client.go
+++ b/pkg/socket/client.go
@@ -3,6 +3,7 @@ package socket
 import (
 	"bufio"
 	"crypto/tls"
+	"errors"
 	"net"
 	"time"
 )
@@ -20,30 +21,32 @@ type socket struct {
 }
 
 func (s *socket) Connect() (err error) {
-	if s.TlsConf == nil {
+	if s.TLSConf == nil {
 		s.conn, err = net.DialTimeout(string(s.Network), s.Address, s.Timeout)
 	} else {
 		var d net.Dialer
 		d.Timeout = s.Timeout
-		s.conn, err = tls.DialWithDialer(&d, string(s.Network), s.Address, s.TlsConf)
+		s.conn, err = tls.DialWithDialer(&d, string(s.Network), s.Address, s.TLSConf)
 	}
 	return err
 }
 
-func (s *socket) Disconnect() error {
-	err := s.conn.Close()
-	s.conn = nil
+func (s *socket) Disconnect() (err error) {
+	if s.conn != nil {
+		err = s.conn.Close()
+		s.conn = nil
+	}
 	return err
 }
 
 func (s *socket) Command(command string, process Processor) error {
-	if err := s.send(command, s.conn, s.Timeout); err != nil {
+	if err := write(command, s.conn, s.Timeout); err != nil {
 		return err
 	}
 	return read(s.conn, process, s.Timeout)
 }
 
-func (s *socket) send(command string, writer net.Conn, timeout time.Duration) error {
+func write(command string, writer net.Conn, timeout time.Duration) error {
 	if err := writer.SetWriteDeadline(time.Now().Add(timeout)); err != nil {
 		return err
 	}
@@ -52,6 +55,9 @@ func (s *socket) send(command string, writer net.Conn, timeout time.Duration) er
 }
 
 func read(reader net.Conn, process Processor, timeout time.Duration) error {
+	if process == nil {
+		return errors.New("process func is nil")
+	}
 	if err := reader.SetReadDeadline(time.Now().Add(timeout)); err != nil {
 		return err
 	}
--- a/pkg/socket/client.go
+++ b/pkg/socket/client.go
@@ -57,6 +57,9 @@ func (s *Socket) Disconnect() (err error) {
 // of the responses this function will stop processing and return a
 // timeout error.
 func (s *Socket) Command(command string, process Processor) error {
+	if s.conn == nil {
+		return errors.New("cannot send command on nil connection")
+	}
 	if err := write(command, s.conn, s.Timeout); err != nil {
 		return err
 	}
@@ -64,6 +67,9 @@ func (s *Socket) Command(command string, process Processor) error {
 }
 
 func write(command string, writer net.Conn, timeout time.Duration) error {
+	if writer == nil {
+		return errors.New("attempt to write on nil connection")
+	}
 	if err := writer.SetWriteDeadline(time.Now().Add(timeout)); err != nil {
 		return err
 	}
@@ -74,6 +80,9 @@ func write(command string, writer net.Conn, timeout time.Duration) error {
 func read(reader net.Conn, process Processor, timeout time.Duration) error {
 	if process == nil {
 		return errors.New("process func is nil")
+	}
+	if reader == nil {
+		return errors.New("attempt to read on nil connection")
 	}
 	if err := reader.SetReadDeadline(time.Now().Add(timeout)); err != nil {
 		return err
--- a/pkg/socket/client.go
+++ b/pkg/socket/client.go
@@ -1,0 +1,64 @@
+package socket
+
+import (
+	"bufio"
+	"crypto/tls"
+	"net"
+	"time"
+)
+
+func NewSocket(config Config) *socket {
+	return &socket{
+		Config: config,
+		conn:   nil,
+	}
+}
+
+type socket struct {
+	Config
+	conn net.Conn
+}
+
+func (s *socket) Connect() (err error) {
+	if s.TlsConf == nil {
+		s.conn, err = net.DialTimeout(string(s.Network), s.Address, s.Timeout)
+	} else {
+		var d net.Dialer
+		d.Timeout = s.Timeout
+		s.conn, err = tls.DialWithDialer(&d, string(s.Network), s.Address, s.TlsConf)
+	}
+	return err
+}
+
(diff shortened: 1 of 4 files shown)


pkg/socket/client.go
```diff
@@ -0,0 +1,62 @@
...
+		s.conn, err = tls.DialWithDialer(&d, string(s.Network), s.Address, s.TlsConf)
+	}
+	return err
+}
+
+func (s *socket) Disconnect() error {
+	err := s.conn.Close()
+	s.conn = nil
+	return err
+}
```
ilyam8, line 37: This method panics if "s.conn" is nil. I think that we shouldn't do it.

pkg/socket/client.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (s *socket) Command(command string, process Processor) error {
+	if err := s.send(command, s.conn, s.Timeout); err != nil {
+		return err
+	}
+	return read(s.conn, process, s.Timeout)
+}
+
+func (s *socket) send(command string, writer net.Conn, timeout time.Duration) error {
+	if err := writer.SetWriteDeadline(time.Now().Add(timeout)); err != nil {
```
ilyam8, line 47: Let's rename this method to "write". Also, it can be a function (I see we don't really use "*socket" there).

pkg/socket/client.go
```diff
@@ -0,0 +1,62 @@
...
+		return err
+	}
+	_, err := writer.Write([]byte(command))
+	return err
+}
+
+func read(reader net.Conn, process Processor, timeout time.Duration) error {
+	if err := reader.SetReadDeadline(time.Now().Add(timeout)); err != nil {
+		return err
```
ilyam8, line 56: This method panics if "process" is nil. I think we shouldn't do it.

pkg/socket/client.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (s *socket) Disconnect() error {
+	err := s.conn.Close()
+	s.conn = nil
+	return err
+}
+
+func (s *socket) Command(command string, process Processor) error {
+	if err := s.send(command, s.conn, s.Timeout); err != nil {
```
ilyam8, line 40: Panics if "process" is nil.

pkg/socket/types.go
```diff
@@ -0,0 +1,30 @@
...
+	Command(command string, process Processor) error
+}
+
+type Processor func([]byte) bool
+
+type Config struct {
+	Network Network
+	Address string
+	Timeout time.Duration
+	TlsConf *tls.Config
```
ilyam8, line 20: Let's make it TLSConf

pkg/socket/types.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+type Network string
+
+const (
+	NetworkIP   Network = "tcp"
+	NetworkTCP  Network = "tcp"
+	NetworkUDP  Network = "udp"
+	NetworkUnix Network = "unix"
+)
```
ilyam8, line 30: That is not used anywhere. Why do we need it?

pkg/socket/client.go
```diff
@@ -0,0 +1,62 @@
...
+	"net"
+	"time"
+)
+
+func NewSocket(config Config) *socket {
+	return &socket{
+		Config: config,
+		conn:   nil,
+	}
+}
```
ilyam8, line 15: I think we need either return public method or interface. This won't do if we add documentation - we can documents what "socket" is.

pkg/socket/types.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"crypto/tls"
+	"time"
+)
+
+type Socket interface {
+	Connect() error
+	Disconnect() error
+	Command(command string, process Processor) error
```
ilyam8, line 11: Why do we need it now? Do we plan to have several implementations of "Socket"?

ilyam8:
@georgeok thanks for the work 👍 One more thing - I noticed that the package is not documented. Is that because it is wip?

georgeok:
> @georgeok thanks for the work 👍 One more thing - I noticed that the package is not documented. Is that because it is wip?

Yes correct. I was waiting for some initial feedback before finalising the changes and adding documentation.

pkg/socket/types.go
georgeok: This with help us with the unit tests once we use the package in a collector. The implementation could just create a mock socket in order to mimic errors rather than implementing a TCP or a UDP server. At the moment in the module we don't need it but it should become handy on the first implementation. I can remove it, but I think it's going to be useful quickly.

pkg/socket/client.go
georgeok: done

pkg/socket/client.go
georgeok: done

pkg/socket/client.go
georgeok: fixed

pkg/socket/client.go
georgeok: fixed again

pkg/socket/types.go
georgeok: done

pkg/socket/types.go
georgeok: This was my mistake. I use them in the tests now. The idea is to give to the implementer the available options. Something similar to what the `http` packages does with the status codes `http.StatusAccepted` etc so that they know what is supported rather than passing strings directly. The is also a bug `NetworkIP   Network = "tcp"` which became `NetworkIP   Network = "ip" but I might do some work here in order to support plain ip transactions.

pkg/socket/types.go
georgeok: So, having worked a little bit more on this, I have removed the interface. We can mock the connection in the `Socket` struct for testing. I believe this is also working better with the provided documentation (previous comment).

pkg/socket/client.go
georgeok: Ok, so I've made Socket in order to provide documentation to the users.

pkg/socket/types.go
georgeok: hmm, the thing that I forgot is that IP packets need raw sockets and we will get permission errors so I removed the `IP` option for the time beiing.

pkg/socket/client.go
```diff
@@ -0,0 +1,85 @@
...
+// timeout value from the Socket config and returns read, write and
+// timeout errors if any. If a timeout occurs during the processing
+// of the responses this function will stop processing and return a
+// timeout error.
+func (s *Socket) Command(command string, process Processor) error {
+	if err := write(command, s.conn, s.Timeout); err != nil {
+		return err
+	}
+	return read(s.conn, process, s.Timeout)
+}
```
ilyam8, line 64: How do we handle closed connections? Or situations when we need to recreate the connection internally in order to write/send?

Also, this function panics if we don't call Connect(). If that is expected, I think we need to mention it in the description.

ilyam8:
> when we do "Connect; Command; Close"

But if all modules follow this pattern I think we are fine with what we have 🤔

georgeok:
Hmm... Some thoughts on the last question (keeping the connection alive):
- we need to be able to have an infinity timeout (or at least a big value). The architecture of the `Command` function will need to change and become async with channels for cancelation. 
- if we need to reconnect we will get back a net.Error (I think). We can check something `net.AddrError.Temporary()` or is `net.AddrError.Timeout()` to see if it's temporary error.
- Have another config param `automatic retry connections` or return the temporary flag to the caller and let him decide to reconnect or not.

ilyam8:
@georgeok thanks for the work! Let's try to use this package and see if there are any problems.

PR #627, opened 2021-09-28
zookeeper use pkg/socket

georgeok:


ilyam8 (approved):

ilyam8:
@jrevillard it seems you have an old version of go. Also, we need to get rid of using my hashsctucure fork.

Do you have docker on your system?

PR #639, opened 2021-12-03
fetch mongodb replica set stats

### LINKED ISSUE #634: mongodb replica set stats not working
I'm running MongoDB 5.1.0 and I'm not seeing the go.d mongodb collector pulling replication stats in either the dashboard or when running `./go.d.plugin -d -m mongodb`

I've tested that I can run `db.serverStatus().repl` in the mongosh cli shell using the same mongodb user, which has the following permissions when I run `use admin` and `db.getUser('monitor')` in mongosh:
```
user: 'monitor',
db: 'admin',
roles: [
  { role: 'clusterMonitor', db: 'admin' },
  { role: 'read', db: 'local' },
  { role: 'read', db: 'admin' }
]
```

I'm using this config in `go.d/mongodb.conf`:
```
jobs:
  - name: local
    uri: 'mongodb://monitor:xxxxx@127.0.0.1:27017/?authSource=admin'
    timeout: 20
```

Based on release notes and research, this was implemented in #604 and should be included in the latest version.

I'm running:
`netdata v1.32.0-1-nightly`
`go.d.plugin, version: v0.31.0`

bompus:
fixes #634

### DIFF
--- a/modules/mongodb/connector.go
+++ b/modules/mongodb/connector.go
@@ -43,7 +43,7 @@ type mongoCollector struct {
 // `serverStatus` command.
 func (m *mongoCollector) serverStatus() (*serverStatus, error) {
 	var status *serverStatus
-	command := bson.D{{Key: "serverStatus", Value: 1}, {Key: "metrics", Value: 0}, {Key: "repl", Value: 0}}
+	command := bson.D{{Key: "serverStatus", Value: 1}, {Key: "metrics", Value: 0}, {Key: "repl", Value: 1}}
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*m.Timeout)
 	defer cancel()
 	err := m.Client.Database("admin").RunCommand(ctx, command).Decode(&status)


modules/mongodb/connector.go
```diff
@@ -43,7 +43,7 @@ type mongoCollector struct {
 // `serverStatus` command.
 func (m *mongoCollector) serverStatus() (*serverStatus, error) {
 	var status *serverStatus
-	command := bson.D{{Key: "serverStatus", Value: 1}, {Key: "metrics", Value: 0}, {Key: "repl", Value: 0}}
+	command := bson.D{{Key: "serverStatus", Value: 1}, {Key: "metrics", Value: 0}, {Key: "repl", Value: 1}}
```
ilyam8, line 46: @georgeok do you remember why we set it to 0? I remember we discussed something like if it is 1 it returns **much** more metrics? Or it was another request?

ilyam8:
@bompus big thanks for finding and fixing the problem 👍

ilyam8 (approved):

PR #644, opened 2022-02-02
WIP: initial draft SNMP collector for go.d.plugin

codeguru1:


modules/snmp/collect.go
```diff
@@ -0,0 +1,44 @@
+package snmp
+
+import (
+	"github.com/gosnmp/gosnmp"
+)
+
+func (s *SNMP) collect() (map[string]int64, error) {
+	collected := make(map[string]int64)
+	var all_oid []string
```
ilyam8, line 9: Let's use camelCase.

modules/snmp/collect.go
```diff
@@ -0,0 +1,44 @@
...
+	return collected, nil
+}
+
+func (s *SNMP) collectChart(collected map[string]int64, oid_s []string) error {
+	params := s.Config.SNMPClient
+	if len(oid_s) > s.Config.MaxOIDs {
+		if err := s.collectChart(collected, oid_s[s.Config.MaxOIDs:]); err != nil {
+			return err
+		}
+		oid_s = oid_s[:s.Config.MaxOIDs]
```
ilyam8, line 31: Please camelCase

modules/snmp/validators.go
```diff
@@ -0,0 +1,144 @@
...
+		err = appendError(err, fmt.Sprintf("invalid or missing value: charts[%d].dimension[%d].name;", index_chart, index))
+	}
+	if d.OID == "" {
+		err = appendError(err, fmt.Sprintf("missing value: charts[%d].dimension[%d].oid;", index_chart, index))
+	}
+	if d.Algorithm == "" ||
+		(d.Algorithm != string(module.Incremental) &&
+			d.Algorithm != string(module.PercentOfIncremental) &&
+			d.Algorithm != string(module.PercentOfAbsolute)) {
+		err = appendError(err, fmt.Sprintf("invalid or missing value: charts[%d].dimension[%d].algorithm;", index_chart, index))
```
ilyam8, line 39: An empty value is ok here (Netdata will use an internal default), it is better to have as many optional parameters as possible.

modules/snmp/validators.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if d.Multiplier == 0 {
+		err = appendError(err, fmt.Sprintf("integer set to 0: charts[%d].dimension[%d].multiplier;", index_chart, index))
+	}
```
ilyam8, line 43: An empty/unset value is ok here (Netdata will use an internal default), it is better to have as less mandatory parameters as possible. 0 means no multiplier.

modules/snmp/validators.go
```diff
@@ -0,0 +1,144 @@
...
+	if d.Divisor == 0 {
+		err = appendError(err, fmt.Sprintf("integer set to 0: charts[%d].dimension[%d].divisor;", index_chart, index))
+	}
```
ilyam8, line 46: An empty/unset value is ok here (Netdata will use an internal default), it is better to have as less mandatory parameters as possible. 0 means no divisor.

modules/snmp/collect.go
codeguru1: ok

modules/snmp/collect.go
codeguru1: done

config/go.d/snmp.conf
```diff
@@ -0,0 +1,168 @@
...
+             { name: "out",
+                oid: "1.3.6.1.2.1.2.2.1.20.2",
+                algorithm: "incremental",
+                multiplier: -8,
+                divisor: 1024,
+              },
+            ]
+        }
+      - { title: "title 2",
+          priority: 2
```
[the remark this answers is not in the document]
codeguru1: sure

modules/snmp/collect.go
ilyam8: Thanks! If something is done feel free to "Resolve conversation" after pushing a commit(s)

modules/snmp/validators.go
codeguru1: I think this check should still be valid. We can initialize multiplier=1 for unset value at init(). But if user exclusively sets it to 0, then it should return error.

modules/snmp/validators.go
codeguru1: same as multiplier comment

modules/snmp/validators.go
codeguru1: The approach here is that, validator shall be strict. But any unset/default values must be handled at init() function.

ilyam8: Mmm, what I mean is an empty (missing value) "Algorithm" is not an error/problem. It is valid.

modules/snmp/validators.go
ilyam8: 0 is the zero value for `int`. Does it mean this check will always fail if not explicitly set?

codeguru1: I mean, if we need to set default, it should happen before validating. Validator is a way to verify even the default values set by us. So I can check if we can initialize default at some other places like init() function or introduce default struct tags.

ilyam8:
It is not really finished, not implemented features and issues will be addressed in subsequent PRs.

PR #653, opened 2022-03-09
fix segfault issue while closing connection

codeguru1:


modules/snmp/snmp.go
```diff
@@ -157,5 +157,8 @@ func (s *SNMP) Collect() map[string]int64 {
 }
 
 func (s SNMP) Cleanup() {
-	s.SNMPClient.Close()
+	//avoid segfault while closing
+	if s.SNMPClient.Target() == s.Name {
+		s.SNMPClient.Close()
+	}
 }
```
ilyam8, line 164: Should we check the nil value too?

```go
if s.SNMPClient != nil && ...
```

ilyam8: And can you explain `s.SNMPClient.Target() == s.Name` case? What does it mean?

codeguru1: we don't need to check for `s.SNMPClient != nil`, it will always have a SNMP object with default values. I realize, the second line `s.SNMPClient.Target() == s.Name` is also not right way. The gosnmp library has some serious limitation with the Close() operation. I will figure out a proper way. Ignore this for now.

ilyam8: >  it will always have a SNMP object with default values

Ah, you are right. Somehow I missed we set it in New().

modules/snmp/snmp.go
```diff
@@ -157,5 +167,7 @@ func (s *SNMP) Collect() map[string]int64 {
...
 
 func (s SNMP) Cleanup() {
```
ilyam8, line 169:
```suggestion
func (s *SNMP) Cleanup() {
```

Should we use pointer receiver in this method?

ilyam8 (approved):

Issue #648, opened 2022-02-09
SNMP collector migration


ilyam8:
The plan is to rewrite [node.d/snmp](https://github.com/netdata/netdata/tree/master/collectors/node.d.plugin/snmp) collector. The initial goal is to have the same functionality so we can deprecate nodejs version.

**Feature parity**:

 - [x] SNMP v1, v2c and v3 support.
 - [x] Multiple SNMP devices per job.
 - [x] Each SNMP device can be used to collect data for multiple charts.
 - [x] Each SNMP device will accept one or more batches to report values (can be set using `max_request_size` per SNMP server, to control the size of batches). Node.d version sends up to `max_request_size` OIDs in one SNMP GET request.
 - [x] `multiply_range` configuration option support. Node.d version dynamically creates charts for a device adding indexes from `multiply_range` to every OID. See the [Configuration](https://github.com/netdata/netdata/tree/master/collectors/node.d.plugin/snmp#configuration) section for details.

**What we need to change**: 

- configuration file format should be YAML, not JSON.
- "offset" option is not needed (not 100% sure, but that is my understanding from reading issues/checking code).

ktsaou:
@ilyam8 auto-discovery would also be nice. But I never dived into it. I don't know how it works.

ilyam8:
@ktsaou you are right. That is the first feature for @codeguru1 and our main goal now is to implement the same functionality first.

ktsaou:
> @ktsaou you are right. That is the first feature for @codeguru1 and our main goal now is to implement the same functionality first.

Great! I would love to have an overview of how this works. So, once you master it, please share this knowledge (even with our users).

ilyam8:
@codeguru1 the following can be implemented as a separate PRs, I will update the list:
 - [ ] move the initialization logic to init.go (so we call just call functions in snmp.go Init() (e.x. validateConfig())).
 - [ ] disable gosnmp library logging completely (or enable only in the debug mode)/

codeguru1:
thanks for the testing and comments

codeguru1:
Some PRs related to this feature, as requested by @cpipilas 
https://github.com/netdata/go.d.plugin/pull/654
https://github.com/netdata/go.d.plugin/pull/653
https://github.com/netdata/go.d.plugin/pull/644

PR #654, opened 2022-03-09
Implement charts multiplier for SNMP collector

codeguru1:


modules/snmp/snmp.go
```diff
@@ -68,7 +68,7 @@ type (
 		Type          *string     `yaml:"type,omitempty"`
 		Family        *string     `yaml:"family,omitempty"`
 		MultiplyRange []int       `yaml:"multiply_range,omitempty"`
```
ilyam8, line 70: We don't need `omitempty` tag.

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

modules/snmp/charts.go
```diff
@@ -32,48 +32,65 @@ var defaultDims = module.Dims{
...
-func newChart(chartIn []ChartsConfig) *module.Charts {
-	charts := &module.Charts{}
-	for i, s := range chartIn {
-		c := defaultSNMPchart.Copy()
-		c.ID = fmt.Sprintf(c.ID, i)
-		c.Title = fmt.Sprintf(c.Title, s.Title)
-		if s.Family != nil {
-			c.Fam = *s.Family
-		}
+func newChart(id, multiplier int, s ChartsConfig) *module.Chart {
```
ilyam8, line 35: Let's rename `multiplier` to `oidIndex`

modules/snmp/charts.go
```diff
@@ -32,48 +32,65 @@ var defaultDims = module.Dims{
...
+	c := defaultSNMPchart.Copy()
+	c.ID = fmt.Sprintf(c.ID, id)
+	c.Title = fmt.Sprintf(c.Title, s.Title)
 
-		if s.Units != nil {
-			c.Units = *s.Units
-		}
+	if multiplier != 0 {
```
ilyam8, line 40: `0` is a valid oid index.

modules/snmp/charts.go
```diff
@@ -32,48 +32,65 @@ var defaultDims = module.Dims{
...
+	}
+
+	if s.Type != nil {
+		c.Type = module.ChartType(*s.Type)
+	}
+
+	c.Priority = s.Priority
+	for _, d := range s.Dimensions {
+		oid := d.OID
+		if multiplier != 0 {
```
ilyam8, line 60: 0 is a valid oid index.

modules/snmp/charts.go
```diff
@@ -32,48 +32,65 @@ var defaultDims = module.Dims{
...
+	charts := &module.Charts{}
+	for i, s := range chartIn {
+		if s.MultiplyRange != nil {
+			for j := s.MultiplyRange[0]; j <= s.MultiplyRange[1]; j++ {
+				chart := newChart(i, j, s)
+				_ = charts.Add(chart)
+			}
+		} else {
+			chart := newChart(i, 0, s)
+			_ = charts.Add(chart)
```
ilyam8, line 92: Let's don't ignore an error

modules/snmp/validators.go
```diff
@@ -77,14 +77,25 @@ func (c ChartsConfig) validateConfig(index_chart int) error {
...
 	}
 
+	if c.MultiplyRange != nil {
+		if len(c.MultiplyRange) != 2 {
+			err = appendError(err, fmt.Sprintf("invalid range: charts[%d].multiply_range;", index_chart))
+		} else {
+			if c.MultiplyRange[0] >= c.MultiplyRange[1] {
+				err = appendError(err, fmt.Sprintf("invalid range: charts[%d].multiply_range;", index_chart))
+			}
+		}
```
ilyam8, line 97: What about the case when one or both values are negative?

modules/snmp/snmp.go
codeguru1: agree, fixed in separate PR

ilyam8: perfect, thanks!

modules/snmp/snmp.go
```diff
@@ -135,13 +135,21 @@ func (s *SNMP) Init() bool {
...
-		_ = c.AddDim(defaultDims[0])
-		_ = c.AddDim(defaultDims[1])
+		for _, d := range defaultDims {
+			if err = c.AddDim(d); err != nil {
+				s.Errorf("Population of charts failed: %v", err)
+				return false
+			}
+		}
 		s.charts = &module.Charts{c}
 	}
```
ilyam8, line 154: I don't think we need that `else` case with the default chart. The assumption that we need to collect ifInOctets/ifOutOctets for interface with index 2 (why 2?) looks pretty questionable to me.

ilyam8: I see that it is not a new code. So we can merge it as is, but let's remove this case right away in the following PR.

ilyam8: Perhaps I am missing something, if so please elaborate on why you think we need that.

config/go.d/snmp.conf
```diff
@@ -162,22 +163,19 @@ jobs:
...
         family: "lan"
-        multiply_range: [1,24]
+        multiply_range: [1,5]
         dimensions:
           - name: "in"
-            oid: "1.3.6.1.2.1.2.2.1.10.2"
+            oid: ".1.3.6.1.2.1.2.2.1.10"
             algorithm: "incremental"
             multiplier: 8
             divisor: 1024
```
ilyam8, line 177: The correct divisor is 1000 when converting bits to kilobits.

modules/snmp/snmp.go
ilyam8: Ok, added it to https://github.com/netdata/go.d.plugin/issues/648#issuecomment-1057198748

ilyam8 (approved):

PR #655, opened 2022-03-10
refactor init function

codeguru1:


modules/snmp/snmp.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/netdata/go.d.plugin/agent/module"
 )
 
+var SNMPHandler = gosnmp.NewHandler
```
ilyam8, line 11: Let's make it private. I think we don't need it public.

ilyam8: Ok, we can do it in one of following PRs, no problem.

ilyam8 (approved):

PR #659, opened 2022-03-14
Remove default charts and make it mandatory parameter

codeguru1:


modules/snmp/snmp.go
```diff
@@ -95,23 +93,10 @@ func (s *SNMP) Init() bool {
...
-		c.ID = fmt.Sprintf(c.ID, 1)
-		c.Title = fmt.Sprint(c.Title, "default")
-		for _, d := range defaultDims {
-			if err = c.AddDim(d); err != nil {
-				s.Errorf("Population of charts failed: %v", err)
-				return false
-			}
-		}
-		s.charts = &module.Charts{c}
+	s.charts, err = allCharts(s.ChartInput)
```
ilyam8, line 96: Let's rename allCharts to newCharts

modules/snmp/config_validator.go
```diff
@@ -124,6 +124,16 @@ func (o Options) validateConfig() error {
 func (s SNMP) validateConfig() error {
 	var err error
 	err = nil
+	if s.ChartInput == nil {
+		err = appendError(err, "charts missing from config")
+	} else {
+		for i, chart_in := range s.ChartInput {
+			if e := chart_in.validateConfig(i); e != nil {
```
ilyam8, line 131: Let's use camelCase for chart_in

modules/snmp/snmp.go
codeguru1: we already have a function name `newChart()`. having another functions `newCharts()` at a glance is pretty confusing. `allCharts()` isn't a bad name afterall for this situation. Let's keep it this way

ilyam8: But you can keep it as it is if you want, no problem.

ilyam8 (approved):

PR #661, opened 2022-03-15
SNMP: check for non-existent OIDs during collect()

codeguru1:
This commit checks for the return values during oid Get() and removes the gosnmp logger

### DIFF
--- a/modules/snmp/collect.go
+++ b/modules/snmp/collect.go
@@ -38,7 +38,13 @@ func (s *SNMP) collectChart(collected map[string]int64, OIDs []string) error {
 	}
 
 	for i, oid := range OIDs {
-		collected[oid] = gosnmp.ToBigInt(result.Variables[i].Value).Int64()
+		switch result.Variables[i].Type {
+		case gosnmp.NoSuchInstance:
+		case gosnmp.NoSuchObject:
+			continue
+		default:
+			collected[oid] = gosnmp.ToBigInt(result.Variables[i].Value).Int64()
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


modules/snmp/collect.go
```diff
@@ -38,7 +38,13 @@ func (s *SNMP) collectChart(collected map[string]int64, OIDs []string) error {
...
 
 	for i, oid := range OIDs {
-		collected[oid] = gosnmp.ToBigInt(result.Variables[i].Value).Int64()
+		switch result.Variables[i].Type {
+		case gosnmp.NoSuchInstance:
+		case gosnmp.NoSuchObject:
+			continue
+		default:
+			collected[oid] = gosnmp.ToBigInt(result.Variables[i].Value).Int64()
+		}
```
ilyam8, line 47:
```suggestion
		switch result.Variables[i].Type {
		case gosnmp.NoSuchInstance:
		case gosnmp.NoSuchObject:
		default:
			collected[oid] = gosnmp.ToBigInt(result.Variables[i].Value).Int64()
		}
```

We don't really need `continue` here

---

@codeguru1 can we have a test case for different types?

codeguru1: I'm working on test cases

modules/snmp/collect.go
```diff
@@ -39,8 +39,8 @@ func (s *SNMP) collectChart(collected map[string]int64, OIDs []string) error {
 
 	for i, oid := range OIDs {
 		switch result.Variables[i].Type {
-		case gosnmp.NoSuchInstance:
-		case gosnmp.NoSuchObject:
+		case gosnmp.NoSuchInstance, gosnmp.NoSuchObject:
+			s.Warningf("Skipping OID %s, no such object", OIDs[i])
```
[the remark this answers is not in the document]
ilyam8: and you can use `oid` instead of `OIDs[i]`

ilyam8 (approved):

PR #662, opened 2022-03-16
SNMP: this commit fixes auth/priv flags mismatch

codeguru1:
Add more options to auth_proto, priv_proto.
Fix auth/priv flags mismatch
Update snmp.conf accordingly

### DIFF
--- a/modules/snmp/init.go
+++ b/modules/snmp/init.go
@@ -19,7 +19,7 @@ func generateSNMPv3MsgFlags(level int) gosnmp.SnmpV3MsgFlags {
 	case 3:
 		flag = gosnmp.AuthPriv
 	}
-	return gosnmp.SnmpV3MsgFlags(flag)
+	return flag
 }
 
 func (s SNMP) initSNMPClient() (gosnmp.Handler, error) {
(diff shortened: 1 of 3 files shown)


modules/snmp/init.go
```diff
@@ -9,6 +9,19 @@ import (
 
 var snmpHandler = gosnmp.NewHandler
 
+func generateSNMPv3MsgFlags(level int) gosnmp.SnmpV3MsgFlags {
+	var flag gosnmp.SnmpV3MsgFlags
```
ilyam8, line 13:
```suggestion
	flag := gosnmp.NoAuthNoPriv
```

Perhaps make it more explicit?

modules/snmp/init.go
```diff
@@ -9,6 +9,19 @@ import (
...
+	switch level {
+	case 1:
+		flag = gosnmp.NoAuthNoPriv
+	case 2:
+		flag = gosnmp.AuthNoPriv
+	case 3:
+		flag = gosnmp.AuthPriv
+	}
+	return gosnmp.SnmpV3MsgFlags(flag)
```
ilyam8, line 22:
```suggestion
	return flag
```

I think we don't need this cast.

ilyam8 (approved):

PR #664, opened 2022-03-20
fix(snmp): assorted cleanup and fixes

ilyam8:


modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
             divisor: 1000
 ```
 
-`update_every` is the update frequency for each server, in seconds.
-
-`family` sets the name of the submenu of the dashboard each chart will appear under.
+### Example: using SNMPv3
 
-`multiplier` and `divisor` are passed by the plugin to the Netdata daemon and are applied to the metric to convert it properly to `units`. 
+To use SNMPv3:
```
kickoke, line 119:
```suggestion
#### Requirements
To use SNMPv3:
```

modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
+Edit the `go.d/snmp.conf` configuration file using `edit-config` from the
+Netdata [config directory](https://learn.netdata.cloud/docs/configure/nodes#the-netdata-config-directory), which is
+typically at `/etc/netdata`.
+
+```bash
+cd /etc/netdata # Replace this path with your Netdata config directory
+sudo ./edit-config go.d/snmp.conf
+```
+
+| Parameter                    | Default value  | Description                                                                                                      |
```
kickoke, line 32: First of all: I LOVE the table! 

Regarding document structure: 
I would place it at the end of the document, since it is a reference. We start a new section 
`### Configuration parameters` where the table lives.

modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
-Create a config file `/etc/netdata/go.d/snmp.conf`. Following is an example of SNMPv3 config.
+
+Edit the `go.d/snmp.conf` configuration file using `edit-config` from the
+Netdata [config directory](https://learn.netdata.cloud/docs/configure/nodes#the-netdata-config-directory), which is
+typically at `/etc/netdata`.
+
+```bash
+cd /etc/netdata # Replace this path with your Netdata config directory
+sudo ./edit-config go.d/snmp.conf
+```
```
kickoke, line 30: I will leave this part as is right now, but will be updated with a component in a future iteration.

modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
+| charts.multiply_range        |       []       | is used when you need to define many charts [using incremental OIDs](#example-using-chartsmultiply_range-option) |
+| charts.dimensions            |       []       | the list of chart dimensions                                                                                     |
+| charts.dimensions.oid        |       -        | the OID path to the metric you [want to collect](#finding-oids)                                                  |
+| charts.dimensions.name       |       -        | the name of the dimension as it will appear at the legend of the chart                                           |
+| charts.dimensions.algorithm  |    absolute    | the dimension algorithm (one of absolute, incremental)                                                           |
+| charts.dimensions.multiplier |       1        | the value to multiply the collected value, applied to convert it properly to units                               |
+| charts.dimensions.divisor    |       1        | the value to divide the collected value, applied to convert it properly to units                                 |
+
+### Example: using SNMPv1/2
```
kickoke, line 64:
```suggestion
### Example: Using SNMPv1/2
```

modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
+          - name: "out"
+            oid: "1.3.6.1.2.1.2.2.1.16.2"
             multiplier: -8
             divisor: 1000
 ```
 
-`update_every` is the update frequency for each server, in seconds.
-
-`family` sets the name of the submenu of the dashboard each chart will appear under.
+### Example: using SNMPv3
```
kickoke, line 117:
```suggestion
### Example: Using SNMPv3
```

modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
 
-`multiplier` and `divisor` are passed by the plugin to the Netdata daemon and are applied to the metric to convert it properly to `units`. 
+To use SNMPv3:
 
+- use `user` instead of `community`.
+- set `options.version` to 3.
 
-If many charts need to be defined using incremental OIDs, `multiply_range` can be used as in following example:
+The rest as in the SNMPv1/2 [example](#example-using-snmpv12).
```
kickoke, line 124:
```suggestion
#### SNMP v3 sample

The rest of the configuration is the same as in the SNMPv1/2 [example](#example-using-snmpv12).
```
If the rest is the same, what does the example below illustrate?

modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
--   "3" is `sha`
--   "4" is `sha224`
--   "5" is `sha256`
--   "6" is `sha384`
--   "7" is `sha512`
+## Data collection speed
 
-Privacy protocols (`user.priv_proto`):
+Keep in mind that many SNMP switches and routers are very slow. They may not be able to report values per second.
+`go.d.plugin` reports the time it took for the SNMP device to respond when executed in the [debug](#troubleshooting)
```
kickoke, line 220:
```suggestion
`go.d.plugin` reports the time it took for the SNMP device to respond when executed in the [debug](#debug)
```

modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
-The above will run it on your console and you will be able to see what Netdata sees, but also errors.
-If it works, restart Netdata to activate the snmp collector and refresh the dashboard (if your SNMP device responds with a delay, you may need to refresh the dashboard in a few seconds).
+- `-t 20` is the timeout in seconds.
+- `-O fn` will display full OIDs in numeric format.
+- `-v 2c` is the SNMP version.
+- `-c public` is the SNMP community.
+- `192.0.2.1` is the SNMP device.
 
-## Data collection speed
+## Troubleshooting
```
kickoke, line 241:
```suggestion
<a id="debug"> <a/>
## Debugging the SNMP collector

```

modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
 
-Keep in mind that many SNMP switches and routers are very slow. They may not be able to report values per second. If `go.d.plugin` is executed in `debug` mode, it will report the time it took for the SNMP device to respond.
+To troubleshoot issues with the `snmp` collector, run the `go.d.plugin` with the debug option enabled. The output should
```
kickoke, line 243: We'll leave this section as is for now. We do have a component planned for debugging, so let's see how well it works for this one. :)

modules/snmp/README.md
```diff
@@ -6,219 +6,254 @@ sidebar_label: "SNMP"
...
--   "2" is `des`
--   "3" is `aes`
--   "4" is `aes192`
--   "5" is `aes256`
+Also, if many SNMP clients are used on the same SNMP device at the same time, values may be skipped. This is a problem
+of the SNMP device, not this collector. In this case, consider reducing the frequency of data collection (
+increasing `update_every`).
 
-## Testing the configuration
+## Finding OIDs
```
kickoke, line 228: Should we add a paragraph on why you should need to find an OID or can we assume our users are familiar with this?

kickoke (commented):
Can do a second review tomorrow, if needed. 
I ran out of concentration 🧠

modules/snmp/README.md
ilyam8: The security parameters needed for SNMPv3 (auth and privacy)

```yaml
    user:
      name: "username"
      level: "authPriv"
      auth_proto: "sha256"
      auth_key: "auth_protocol_passphrase"
      priv_proto: "aes256"
      priv_key: "priv_protocol_passphrase"
```

modules/snmp/README.md
ilyam8: This is just a copy from https://learn.netdata.cloud/docs/agent/collectors/node.d.plugin/snmp#finding-oids

We assume users have a basic understanding of SNMP protocol.

> why you should need to find an OID

Because it is a crucial part of this plugin configuration. "want to collect" is a hyperlink to the section

Screenshot 2022-03-25 at 10 09 34

modules/snmp/README.md
ilyam8: Why do we need "the SNMP collector"? We don't have "Configuring the X collector", but just "Configuration". I think having the same section name in every readme is better. I agree with changing "Troubleshooting" to something else, but this needs to be done in all readmes then.

modules/snmp/README.md
ilyam8: Not sure these are "requirements" there

modules/snmp/README.md
ilyam8: The table loves you back! 😄 

---

> I would place it at the end of the document since it is a reference

I see your point, but I think it can be hard to read/understand the examples w/o going through this table first (also having "Configuration parameters" outside of "Configuration" 🤔 ).

PR #675, opened 2022-04-12
openvpn_status

surajnpn:
Fixes: netdata/netdata#12541


config/go.d.conf
```diff
@@ -77,3 +77,4 @@ modules:
 #  wmi: yes
 #  x509check: yes
 #  zookeeper: yes
+#  openvpn_status: yes
```
thiagoftsm, line 80: please, add a new line at the end of the file.

modules/openvpn_status/openvpn_status.go
```diff
@@ -0,0 +1,96 @@
...
+
+func init() {
+	creator := module.Creator{
+		Defaults: module.Defaults{
+			Disabled: true,
+		},
+		Create: func() module.Module { return New() },
+	}
+
+	module.Register("openvpn_status", creator)
```
ilyam8, line 22: let's rename it to `openvpn_status_log`:
 - (i believe) the default status filename is 'openvpn-status.log'.
 - having in mind that we have `openvpn` collector, adding the `_log` prefix will make it less confusing what is what.

modules/openvpn_status/openvpn_status.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+const (
+	defaultFilePath = "/var/log/openvpn/status.log"
+)
+
+func init() {
+	creator := module.Creator{
+		Defaults: module.Defaults{
+			Disabled: true,
```
ilyam8, line 17: Why disabled?

modules/openvpn_status/openvpn_status_test.go
```diff
@@ -0,0 +1,108 @@
...
+TIME	2022-03-31 11:53:29	1648716809
+HEADER	CLIENT_LIST	Common Name	Real Address	Virtual Address	Virtual IPv6 Address	Bytes Received	Bytes Sent	Connected Since	Connected Since (time_t)	UsernameClient ID	Peer ID	Data Channel Cipher
+CLIENT_LIST	gofle	192.168.122.135:49199	10.8.0.2		19265	261631	2022-03-31 11:53:19	1648716799	UNDEF	0	0	AES-128-GCM
+CLIENT_LIST     client_bsd2     192.168.122.136:21748   10.8.0.3       2001:0db8:85a3:0000:0000:8a2e:0370:7334         2752   3545  2022-04-16 20:48:27     1650131307      UNDEF   624     0       AES-128-GCM
+HEADER	ROUTING_TABLE	Virtual Address	Common Name	Real Address	Last Ref	Last Ref (time_t)
+ROUTING_TABLE	10.8.0.2	gofle	192.168.122.135:49199	2022-03-31 11:53:24	1648716804
+ROUTING_TABLE   10.8.0.3        client_bsd2     192.168.122.136:21748   2022-04-16 20:48:27     1650131307
+GLOBAL_STATS	Max bcast/mcast queue length	0
+END`
+)
```
ilyam8, line 49: Let's have them as txt files under `testdata/`. That is a common pattern we use, also it will make testing easier for you since it will be closer to the real case (reading files).

modules/openvpn_status/openvpn_status_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func TestOpenVPN_Status_Init(t *testing.T) {
+	assert.True(t, New().Init())
+}
+
+func TestOpenVPN_Status_Collect(t *testing.T) {
+	for _, logData := range statusLogList {
+		statusPath, cleanup := createTempFile(logData)
+		defer cleanup()
+		job := New()
```
ilyam8, line 95: I suggest having a test case for a scenario:
 - status 1 w/o clients
 - status 2 w/o clients
 - status 3 w/o clients
 - status 1 w/ clients
 - status 2 w/ clients
 - status 3 w/ clients
 - etc...

ilyam8 (changes requested):

modules/openvpn_status/openvpn_status.go
surajnpn: Until we remove python collector with same name?

ilyam8: The python one name is ovpn_status_log. Ideally, we should name the go version ovpn_status_log too (but I think it is ok to not do it): then only one of them (py/go) can be active at the same time.

ilyam8 (approved):

PR #678, opened 2022-05-01
[WIP]add chrony collect modules

boxjan:
We use go.d.plugin to collect chrony metrics in prod env so long time, and it was work very well, we think is quickly than python.d.

Test case will fix later.

We use sentry to statistics problem, I am not sure the repo will welcome it, so I didn't put it in this PR. If you have any suggest please show it.

### DIFF
--- a/config/go.d.conf
+++ b/config/go.d.conf
@@ -18,6 +18,7 @@ modules:
 #  activemq: yes
 #  apache: yes
 #  bind: yes
+#  chrony: yes
 #  cockroachdb: yes
 #  consul: yes
 #  coredns: yes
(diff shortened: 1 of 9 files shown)


ilyam8:
Hi, @boxjan. Big thanks for the contribution! A small request before we start reviewing it - can you make all Public methods private if they are used only in the chrony package (all except Init, Check, Collect, Cleanup).

boxjan:
Hi, @ilyam8 all Public methods/functions have been change private if they are used only in the chrony package,
about the test case, I think we can't fake a reply based on the current request, otherwise we don't know if chrony's reply can really be processed so all of these test need chrony running, and listen at default port.

boxjan:
So the test case only checked what should not be wrong in most cases

ilyam8:
Hi, @boxjan. Why did you decide to not use the https://github.com/facebook/time/tree/main/ntp/chrony library directly?

boxjan:
In fact, we have been using our collector since November 2020, and we were not aware of the package at the time

PR #681, opened 2022-05-05
openvpn_status_log: add support for openvpn configured with static key

surajnpn:


modules/openvpn_status_log/parser.go
```diff
@@ -162,3 +164,30 @@ func parseStatusLogV3(scanner *bufio.Scanner) []clientInfo {
...
+				i := strings.Split(parts[2], ",")
+				in, _ := strconv.Atoi(i[1])
+				client.BytesReceived = in
+			}
+			if parts[1] == "write" {
+				i := strings.Split(parts[2], ",")
+				out, _ := strconv.Atoi(i[1])
+				client.BytesSent = out
+			}
+		}
```
ilyam8, line 190: I suggest checking `len(parts)` before doing indexing.

```go
if len(parts) < 3 {
  return nil
}
```

PR #683, opened 2022-05-11
openvpn_status_log: check for array outbound errors

surajnpn:


ilyam8:
@surajnpn let's add some samples with unexpected data (including an empty file) to `/testdata` and create new test cases with them

modules/openvpn_status_log/openvpn_status_log_test.go
```diff
@@ -122,3 +119,26 @@ func TestOpenVPN_Status_WithStaticKey_NoClients(t *testing.T) {
...
+	job.StatusPath = logFile
+	require.True(t, job.Init())
+	require.True(t, job.Check())
+
+	mx := job.Collect()
+	assert.Equal(t, 0, len(mx))
+}
+
+func TestOpenVPN_Status_WithInvalidFile(t *testing.T) {
+	logFile := "testdata/invalid_file.txt"
```
ilyam8, line 136: Do you mean a non-existent file?

modules/openvpn_status_log/parser.go
```diff
@@ -98,13 +101,19 @@ func parseStatusLogV2(scanner *bufio.Scanner) []clientInfo {
...
+			return nil
+		}
 		switch parts[0] {
 		case "HEADER":
 		case "END":
 			break
 		default:
 			switch statusType := parts[0]; statusType {
 			case "CLIENT_LIST":
+				if len(parts) < 8 {
```
ilyam8, line 114:
```suggestion
				if len(parts) < 9 {
```

because you access 9th element in `strconv.Atoi(parts[8])`

This is not the only place with such error, please check all if (len) checks

modules/openvpn_status_log/openvpn_status_log_test.go
surajnpn: yes

modules/openvpn_status_log/parser.go
```diff
@@ -128,13 +137,19 @@ func parseStatusLogV3(scanner *bufio.Scanner) []clientInfo {
...
```
surajnpn, line 150:
```suggestion
				if len(parts) < 9 {
```

modules/openvpn_status_log/openvpn_status_log_test.go
ilyam8: can we change the name? 'InvalidFile' sounds to me like a file with unexpected data.

modules/openvpn_status_log/openvpn_status_log_test.go
```diff
@@ -122,3 +119,26 @@ func TestOpenVPN_Status_WithStaticKey_NoClients(t *testing.T) {
...
+	job := New()
+	job.StatusPath = logFile
+	require.True(t, job.Init())
+	require.True(t, job.Check())
+
+	mx := job.Collect()
+	assert.Equal(t, 0, len(mx))
+}
+
+func TestOpenVPN_Status_WithInvalidFile(t *testing.T) {
```
ilyam8, line 135:
```suggestion
func TestOpenVPN_Status_WithNonExistentFile(t *testing.T) {
```

ilyam8 (approved):

PR #711, opened 2022-07-12
weblog: Correctly parse multiple values in nginx upstream_response_time

kralewitz:
This commit changes the way how the upstream_response_time field is
parsed in nginx access logs. The field can contain multiple time values
if the request got processed by multiple upstream servers, or if it
passed through an internal redirect between multiple server groups (i.e.
by using the X-Accel-Redirect header).

Previously, only the first encountered value was processed by weblog
plugin, and it only accounted for the ',' separator. If the field contained
the ':' separator, the value was not parsed correctly. With this commit
applied, both of these cases will be handled correctly. Additionaly, the
value reported will be the sum of times from all upstream servers that
handled the request, i.e. the total time the request spent upstream.

### DIFF
--- a/modules/weblog/logline.go
+++ b/modules/weblog/logline.go
@@ -325,19 +325,30 @@ func (l *logLine) assignReqProcTime(time string) error {
 	return nil
 }
 
+func isUpstreamTimeSeparator(r rune) bool { return r == ',' || r == ':' }
+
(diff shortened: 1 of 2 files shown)


ilyam8:
Thanks, @kralewitz 👍

ilyam8 (approved):

PR #763, opened 2022-08-15
feat(mysql): Only produce binlog and MyISAM charts if they're enabled

snwflake:
Fixes https://github.com/netdata/netdata/issues/13518

MariaDB does not know the global `disabled_storage_engines`, so I omitted the test cases and always enabled charts for MyISAM keys.


modules/mysql/collect.go
```diff
@@ -119,6 +126,22 @@ func hasQCacheMetrics(collected map[string]int64) bool {
 	return ok
 }
 
+func hasMyISAMStorageEngine(collected map[string]int64) bool {
+	if collected["disabled_storage_engines"] == 0 {
+		return true
+	} else {
+		return false
+	}
+}
```
ilyam8, line 135:
```suggestion
func hasMyISAMStorageEngine(collected map[string]int64) bool {
    return collected["disabled_storage_engines"] == 0
}
```

modules/mysql/collect.go
```diff
@@ -119,6 +126,22 @@ func hasQCacheMetrics(collected map[string]int64) bool {
...
+
+func hasBinlogEnabled(collected map[string]int64) bool {
+	if collected["log_bin"] == 1 {
+		return true
+	} else {
+		return false
+	}
+}
```
ilyam8, line 143:
```suggestion
func hasBinlogEnabled(collected map[string]int64) bool {
    return collected["log_bin"] == 1
}
```

modules/mysql/collect_global_vars.go
```diff
@@ -53,6 +59,28 @@ func (m *MySQL) collectGlobalVariables(collected map[string]int64) error {
...
 	return strconv.ParseInt(value, 10, 64)
 }
+
+func convertStorageEngineValue(val string) string {
+	if strings.Contains(val, "MyISAM") {
+		return "1"
+	} else {
+		return "0"
+	}
+}
```
ilyam8, line 78:
```suggestion
func convertStorageEngineValue(val string) string {
	if strings.Contains(val, "MyISAM") {
		return "1"
	}
	return "0"
}
```

modules/mysql/collect_global_vars.go
```diff
@@ -53,6 +59,28 @@ func (m *MySQL) collectGlobalVariables(collected map[string]int64) error {
...
+
+func convertBinlogValue(val string) string {
+	if val == "OFF" {
+		return "0"
+	} else {
+		return "1"
+	}
+}
```
ilyam8, line 86:
```suggestion
func convertBinlogValue(val string) string {
	if val == "OFF" {
		return "0"
	}
	return "1"
}
```

ilyam8 (changes requested):

ilyam8:
@snwflake there are some problems (e.g. lack of a testing case for this fix), but taking into account [your comment](https://github.com/netdata/netdata/issues/13518#issuecomment-1214585460) I think it is ok to merge. I plan to do some refactoring of mysql collector and will add the test case in there.

Thanks for the contribution!

Keep in mind that merging this PR doesn't really fix the problem - we need to release a new version of go.d.plugin which I plan to do later this week (no exact date since I am on vacation).

ilyam8 (approved):

PR #828, opened 2022-09-14
make k8s_* collectors chart priority higher than cgroups

ilyam8:
Fixes: netdata/netdata#13663

Waiting for an [answer](https://github.com/netdata/netdata/issues/13663#issuecomment-1246630763) before merging.


ilyam8:
Ok, merging, will revert if needed.

PR #830, opened 2022-09-14
fix(proxysql): fixes and simplification part

ilyam8:


ilyam8:
@yasharne, ok, I went through the ProxySQL metrics:

- problems with dimension algorithm (absolute/incremental). It is not clear from their [docs](https://proxysql.com/documentation/stats-statistics), I had to check their source code to find out which algorithm to use where.
- problems metrics grouping. Again, not clear from the docs, but the source code contains a (brief) description.

I believe I removed some charts because I don't have more time for proxysql today and I want to merge my fix.

Now at least we can remove the python version 🎉 Possible improvements: go through [ProxySQL Grafana dashboard](https://github.com/ProxySQL/proxysql-grafana-prometheus/blob/main/grafana/provisioning/dashboards/ProxySQL-Host-Statistics.json) - it will help to add/regroup metrics.

PR #909, opened 2022-10-10
Update readme

shyamvalsan:
Update readme with steps to create a netdata user in postgres

### DIFF
--- a/modules/postgres/README.md
+++ b/modules/postgres/README.md
@@ -20,7 +20,7 @@ To find out more about these metrics and why they are important to monitor, read
 - User with granted `pg_monitor`
   or `pg_read_all_stat` [built-in role](https://www.postgresql.org/docs/current/predefined-roles.html).
 
-Without additional configuration Netdata will attempt to use the default postgres user - but a separate netdata user can be created for this purpose. If you have PostgreSQL 10+ and want to use Netdata to monitor statistics normally reserved for superusers grant the netdata user pg_monitor permissions. Some of the advanced metrics also require additional permissions as mentioned in [metrics](https://github.com/shyamvalsan/go.d.plugin/edit/patch-2/modules/postgres/README.md?pr=%2Fnetdata%2Fgo.d.plugin%2Fpull%2F909#metrics).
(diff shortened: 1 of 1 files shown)


modules/postgres/README.md
```diff
@@ -20,6 +20,20 @@ To find out more about these metrics and why they are important to monitor, read
...
+2. `postgres=# create user netdata with password '<PASSWORD>';`
+3. `postgres=# grant pg_monitor to netdata;`
+
+Edit the `go.d/postgres.conf` configuration file using `edit-config` from the Netdata [config directory](https://learn.netdata.cloud/docs/configure/nodes), which is typically at `/etc/netdata` and set the username and password you just defined.
+
+```
+jobs:
+  - name: <jobname>
+    dsn: 'postgres://<username>:<password>@<host>:<port>/<dbname>'
+```
```
ilyam8, line 35: this can be a link to the configuration section

modules/postgres/README.md
```diff
@@ -20,6 +20,20 @@ To find out more about these metrics and why they are important to monitor, read
 - User with granted `pg_monitor`
   or `pg_read_all_stat` [built-in role](https://www.postgresql.org/docs/current/predefined-roles.html).
 
+Without additional configuration Netdata will attempt to use the default postgres user - but a separate netdata user can be created for this purpose. If you have PostgreSQL 10+ and want to use Netdata to monitor statistics normally reserved for superusers grant the netdata user pg_monitor permissions. Some of the advanced metrics also require additional permissions as mentioned in [metrics](https://github.com/netdata/go.d.plugin/tree/master/modules/postgres#metrics).
```
ilyam8, line 23:
```suggestion
Without additional configuration Netdata will attempt to use the default postgres user - but a separate netdata user can be created for this purpose. If you have PostgreSQL 10+ and want to use Netdata to monitor statistics normally reserved for superusers grant the netdata user pg_monitor permissions. Some of the advanced metrics also require additional permissions as mentioned in [metrics](#metrics).
```

modules/postgres/README.md
```diff
@@ -20,6 +20,20 @@ To find out more about these metrics and why they are important to monitor, read
...
+
+1. Login to a psql session as a user with CREATEROLE priviliges
+2. `postgres=# create user netdata with password '<PASSWORD>';`
```
ilyam8, line 26: If we suggest passwordless we can add data collection jobs with netdata user, so it would be: create the user, restart netdata service.

ilyam8: In addition, the can be not a list but a code block a user can copy/paste. See https://github.com/netdata/go.d.plugin/tree/master/modules/mysql#requirements

shyamvalsan: How about now?

ilyam8:
Thanks, @shyamvalsan 👍

PR #931, opened 2022-10-20
WMI CPU frequency metric

thiagoftsm:
This PR is bringing following metrics for WMI module:

| Area | metric |
|--------|----------|
| CPU | Frequency |

Screenshot_20221020_140306


modules/wmi/charts.go
```diff
@@ -109,6 +111,15 @@ var (
 		Type:     module.Stacked,
 		Priority: prioCPUInterrupts,
 	}
+	cpuFrequencyChart = Chart{
+		ID:       "cpu_frequency",
+		Title:    "CPU Frequency",
+		Units:    "MHz",
+		Fam:      "cpu",
+		Ctx:      "wmi.cpu_frequency",
+		Type:     module.Line,
```
ilyam8, line 120:
```suggestion
```

Not needed, module.Line is the default

modules/wmi/charts.go
```diff
@@ -840,6 +854,24 @@ func addDimToCPUInterruptsChart(charts *Charts, coreID string) error {
...
+func addDimToCPUFrequencyChart(charts *Charts, coreID string) error {
+	chart := charts.Get(cpuFrequencyChart.ID)
+	if chart == nil {
+		return fmt.Errorf("chart '%s' is not in charts", cpuFrequencyChart.ID)
+	}
+	dim := &Dim{
+		ID:   fmt.Sprintf("cpu_core_%s_frequency", coreID),
+		Name: "core" + coreID,
+		Algo: module.Incremental,
+		Div:  1000,
```
ilyam8, line 866:
```suggestion
```

Not needed when `stm:"frequency,1,1"`

ilyam8, line 866:
```suggestion
	dim := &Dim{
		ID:   fmt.Sprintf("cpu_core_%s_frequency", coreID),
		Name: "core" + coreID,
```

- Div is not needed when stm:"frequency,1,1".
- the value is the gauge, Algorithm should be Absolute (the default when not set).

ilyam8:
@thiagoftsm does the value change? It is pretty strange all of the cores have 2803 freq

ilyam8:
Then I doubt we need that. Probably that was the reason I didn't add CPU frequncey. @vlvkobal can you check `windows_cpu_core_frequency_mhz` on your win laptop?

cc @shyamvalsan

ilyam8:
@thiagoftsm @shyamvalsan my assumption is that `windows_cpu_core_frequency_mhz` is the CPU core base speed and not the actual frequency.

thiagoftsm:
I am also bringing @cakrit for discussion, because I understood that we should collect data to match what we are collecting on Linux.

ilyam8:
I will reopen if needed. @thiagoftsm this is just common sense. Why do we need to create a chart for CPU base speed?

thiagoftsm:
Let us see others opinion.  :)

ilyam8:
@thiagoftsm please share your reasoning why we need that

thiagoftsm:
> @thiagoftsm please share your reasoning why we need that

you can run commands to verify that you bought a computer with a specific frequency, but you could also take a look on chart. :)

cakrit:
If we can, let's put cpu frequency in the place where it belongs, i.e. here. Same for the other host labels/info. This is a separate, but useful task that I will add to Windows support.

image

PR #953, opened 2022-10-29
WMI Process (APPS)

thiagoftsm:
This PR is bringing process monitoring for `Microsoft` like we have for Linux/BSD/Mac processes. 

We are bringing metrics analogous o metrics we are collecting for Linux/BSD/Mac environments.


ilyam8:
Hi. Let's name them process metrics, not apps. I don't think we try to copy Netdata metric names (which will be renamed anyway).

ilyam8 (approved):

thiagoftsm:
> Hi. Let's name them process metrics, not apps. I don't think we try to copy Netdata metric names (which will be renamed anyway).

I confess I did not like the idea to have two names for the same stored objects, probably this creates confusions for users (ping @shyamvalsan and @cakrit ), but I do not have objections to use `process`

ilyam8:
What "the same"? I am not aware that Netdata collects per-process metrics. Not sure I understand the confusion part too.

thiagoftsm:
> What "the same"? I am not aware that Netdata collects per-process metrics. Not sure I understand the confusion part too.

`Process` and `Apps` are different names for the same objects, software running on computer. In my understanding we should use only one word to name them independent of the operate system `netdata` is running.

ilyam8:
Ok, thanks for the explanation! apps.plugin doesn't collect per-process metrics, do you know that?

thiagoftsm:
> Ok, thanks for the explanation! apps.plugin doesn't collect (expose as charts) per-process metrics, do you know that?

Yes, but we organize metrics per apps groups and one day these windows metrics could be per apps groups too. 

The fact that we do not collect per apps does not mean that we should do this. It is completely possible for users to convert apps group in application.

ilyam8:
> could be per apps groups too

Could be or could not be. Just answering to "the same stored objects" - I don't see that processes == (custom) process groups. In addition, I don't understand why you are actively insisting on some equality on Linux metrics names (the ones we have now in Netdata will be renamed, 100% of them will change their names).

PR #954, opened 2022-10-29
chore(wmi): rename apps to processes plus small fixes

ilyam8:


modules/wmi/charts.go
```diff
@@ -821,49 +807,60 @@ func (w *WMI) updateCharts(mx *metrics) {
...
-		if w.cache.apps[apps.ID] {
+
+	procs := make([]string, 0, len(mx.Processes.procs))
+	for _, proc := range mx.Processes.procs {
+		procs = append(procs, proc.ID)
+	}
+	sort.Slice(procs, func(i, j int) bool { return procs[i] < procs[j] })
+
+	for _, id := range procs {
+		if w.cache.procs[id] {
```
thiagoftsm, line 833: Why did you abbreviate `process` ?

ilyam8: Because I decided to do so - procs/processes. Does it look like a problem for you?

thiagoftsm: No, it is not a problem.

When I brought abbreviation for `apps`, it was because this is the way we work on Linux/BSD/Mac. On the other hand we are not abbreviating `memory` when we could.

I am still trying to understand the pattern. :thinking:

ilyam8: > w.cache.**procs**

That is a private variable. Why does it bother you how I name it?

ilyam8: Or do you talk about something different? I am commenting on the highlighted code.

thiagoftsm: I understand this is an internal variable, but why should it to be the unique with abbreviation when others do not have? :thinking:

PR #955, opened 2022-10-31
WMI service metric

thiagoftsm:
This PR is bringing metrics to monitor Windows Services;

Screenshot_20221031_022225


ilyam8 (changes requested):
- The aggregated charts will be broken.
- Aggregation over time will be broken.

If we encode the state/status as an integer.

ilyam8:
@thiagoftsm I don't think we can merge as it is now because of the aforementioned reasons.

modules/wmi/charts.go
```diff
@@ -69,7 +69,14 @@ const (
 	prioProcessesPoolBytes
 	prioProcessesThreads
 
-	prioServiceState
+	prioServiceStateContinuePending
+	prioServiceStatePausePending
```
ilyam8, line 73: I don't think this will do. My suggestion is:
- a chart per service
- the chart dimensions - states.

thiagoftsm: this will be unusable, because we will have hundred/thousands  of charts.. :/

ilyam8: what you are doing is unusable too

thiagoftsm: We need to do something like we are doing right now with Linux services.

ilyam8: > we will have hundred/thousands of charts

As I said - ND Cloud UI (overview)

ilyam8: > We need to do something like we are doing right now with Linux services.

Not possible. Prove me wrong.

thiagoftsm: Windows has hundred of services, times 12 different status we will have thousands, this is the motive I think this is the safest way to go with Microsoft.

ilyam8: No. The number of charts == the number of services * the number of metrics. It is not a safe/unsafe question - we don't really have a lot of options here.

PR #972, opened 2022-11-06
IIS Metrics

thiagoftsm:
This PR is adding metrics to monitor all IIS sites:


modules/wmi/charts.go
```diff
@@ -598,6 +611,132 @@ var (
...
+		IISErrorChart.Copy(),
+	}
+	IISActiveConnectionChartsChart = module.Chart{
+		ID:       "iis_active_conn",
+		Title:    "Website Active connections",
+		Units:    "connections",
+		Fam:      "IIS",
+		Ctx:      "wmi.iis_active_conn",
+		Type:     module.Stacked,
+		Priority: prioIISActiveConns,
```
ilyam8, line 637: We need a chart per site, not a dimension per site.

ilyam8 (changes requested):
a chart per site, please

ilyam8:
@thiagoftsm on second thought, I have no problem with merging it as it is (a dimension per site). The just means we will need to change it in the future. So you can keep it as it is if you want.

thiagoftsm:
> @thiagoftsm on second thought, I have no problem with merging it as it is (a dimension per site). The just means we will need to change it in the future. So you can keep it as it is if you want.

I am already working to modify it. Until the end of the day we will have per site. :handshake:

modules/wmi/charts.go
thiagoftsm: Done

ilyam8 (approved):

PR #978, opened 2022-11-09
fix(postgres):  protect schema/table names with quote_ident

benoitdm-oslandia:
In the case of specific characters in schema or table names, the request failed with: `ERROR:  relation "schemaname.relname" does not exist`


ilyam8:
@benoitdm-oslandia, thanks!

ilyam8 (approved):

PR #980, opened 2022-11-09
add commented example to pull from node exporter demo

andrewm4894:
Make it easy for users to get started with a working remote prometheus demo node exporter so they can see how it looks in netdata without having to set up a localhost prometheus which is what all the default examples and configs assume. 

https://node.demo.do.prometheus.io/metrics

### DIFF
--- a/config/go.d/prometheus.conf
+++ b/config/go.d/prometheus.conf
@@ -207,6 +207,9 @@ jobs:
   # https://github.com/prometheus/prometheus/wiki/Default-port-allocations
   #  - name: node_exporter_local
   #    url: 'http://127.0.0.1:9100/metrics'
+  # example to pull from node exporter demo
+  #  - name: node_exporter_demo
+  #    url: 'https://node.demo.do.prometheus.io/metrics'
   - name: wireguard_local
     url: 'http://127.0.0.1:9586/metrics'
     expected_prefix: 'wireguard_'


config/go.d/prometheus.conf
```diff
@@ -207,6 +207,9 @@ jobs:
   # https://github.com/prometheus/prometheus/wiki/Default-port-allocations
   #  - name: node_exporter_local
   #    url: 'http://127.0.0.1:9100/metrics'
+  # example to pull from node exporter demo
```
ilyam8, line 210: and you need that why? Can you elaborate in the comment

andrewm4894: added comment to pr desc. 

mainly its to give them a really opbvious of a "real" prometheus endpoint out there in the wild that they can just point at. 

i found all the existing ones sort of assume or suggest you need to install prometheus locally and then can scrape `http://127.0.0.1:9100/metrics` for example. 

this is just to try make it even more obvious that all you need is a url with prom metrics on it by showing a "real" (demo) node exporter. 

had a look at the demo site does look to be a proper one created by the promethus team: https://demo.do.prometheus.io/
https://github.com/prometheus/demo-site

ilyam8: I mean add reasoning to the comment

> example to pull from node exporter demo

andrewm4894: Oh sorry, yep totally will do

ilyam8: > i found all the existing ones sort of assume or suggest you need to install prometheus locally and then can scrape http://127.0.0.1:9100/metrics for example.

Yes, that is the assumption we use everywhere - we do try to discover stuff that runs locally.

ilyam8: > this is just to try make it even more obvious that all you need is a url with prom metrics on it by showing a "real" (demo) node exporter.

If the goal of this PR is to highlight that you can use a remote endpoint and not localhost only then I politely disagree. I doubt such confusion is possible.

We have local/remote examples in the [readme](https://github.com/netdata/go.d.plugin/tree/master/modules/prometheus#configuration). Perhaps you can update it.

andrewm4894: I mean, uncomment some lines in a conf file and then just "see" it working.

andrewm4894: But yeah perhaps just mentioning the example I have here in the readme is enough

ilyam8: > uncomment some lines in a conf file and then just "see" it working.

That makes sense. On the other hand - if a person made it to the config file it is very likely that they not just wondering around but have an endpoint to scrape.

ilyam8: >  I have here in the readme is enough

lgtm, lets do this way

andrewm4894:
closed in favour of https://github.com/netdata/go.d.plugin/pull/982

PR #1039, opened 2022-12-26
Fix replication_delta panel for postgres.

artemsafiyulin:
Currently using an incorrect script to calculate the replication lag.
To compare the work of monitoring the replication lag with different requests, I conducted an experiment:

Emulated an artificial network delay on the replica.
Started the activity generating WALK on the master.
Compared 2 different charts

1. Current implementation in Netdata
netdata_lag

2. The solution proposed in this MR (using the example of grafana).
new_lag

We see that the replication lag was actually at the time of sending data to the replica. All other stages passed almost without delay.
This situation arises due to the fact that netdata considers the replication lag for each stage relative to the last generated WAL record on the master. For correct calculation, it is necessary to consider the difference between the current stage of the WAL path and the previous one (not the initial one).


ilyam8:
Hi, @artemsafiyulin. Thanks for the PR! I see you have a negative (-3,7 MB) replay lag on the screenshot. Is it expected?

artemsafiyulin:
> Hi, @artemsafiyulin. Thanks for the PR! I see you have a negative (-3,7 MB) replay lag on the screenshot. Is it expected?

Hello! Yes, this situation can occur when there is a difference between flush_lsn and replay_lsn. That is, the diff between flushed to disk and replayed into the database.
I also note that the data above are taken with a high artificial load on very weak virtual servers.

ilyam8:
@artemsafiyulin yeah, I understand that that is the difference between `flush_lsn` and `replay_lsn`. I am not a Postgres specialist. The question is: is it correct to assume that these steps are done in order and we can calculate the amount of delay between them?

1. `pg_last_wal_receive_lsn(` <=> `sent_location`
2. `sent_location` <=> `write_location`
3. `write_location` <=> `flush_location`
4. `flush_location` <=> `replay_location`

Or do we just need the diff between `pg_last_wal_receive_lsn()` and `replay_location` and call it replication lag size?

ilyam8:
I am trying to understand a negative replay delta (`pg_wal_lsn_diff(flush_lsn, replay_lsn)`):

- positive: the data is stored on a disk (slave) that is not yet replayed and not accessible to end users.
- negative: how do I interpret it?

artemsafiyulin:
> @artemsafiyulin yeah, I understand that that is the difference between `flush_lsn` and `replay_lsn`. I am not a Postgres specialist. The question is: is it correct to assume that these steps are done in order and we can calculate the amount of delay between them?
> 
> 1. `pg_last_wal_receive_lsn(` <=> `sent_location`
> 2. `sent_location` <=> `write_location`
> 3. `write_location` <=> `flush_location`
> 4. `flush_location` <=> `replay_location`
> 
> Or do we just need the diff between `pg_last_wal_receive_lsn()` and `replay_location` and call it replication lag size?

- We really need to calculate the difference between each step in order to understand at which step the delay occurred.
- The diff between pg_last_wal_receive_lsn() and replay_location is a separate metric (total_leg) that really shows the overall replication lag, but does not give us details.

> I am trying to understand a negative replay delta (`pg_wal_lsn_diff(flush_lsn, replay_lsn)`):
> 
> * positive: the data is stored on a disk (slave) that is not yet replayed and not accessible to end users.
> * negative: how do I interpret it?
> 
> If getting negative values is a timing issue and they are useless, I think we need to change them to 0 (in the query, e.g. using [GREATEST](https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-GREATEST-LEAST)). Can you do that?

*negative: As can be seen from the graphs, we can get negative values for reply_lag. Because at that moment we wrote the data to the OS cache, after which postgres applied the data faster than the OS flushed them from the cache to disk. It plays a role that the database does not work with the disk directly, but through the OS cache.

ilyam8:
> postgres applied the data faster than the OS flushed them from the cache to disk

I was under impression that this is not possible [because](https://www.postgresql.org/docs/current/wal-intro.html)

> WAL's central concept is that changes to data files (where tables and indexes reside) must be written only after those changes have been logged, that is, **after log records describing the changes have been flushed to permanent storage**.

Write and flush locations difference -  in case Postgres crashes these changes will be lost. That makes me think that there is no replay until flush.

> plays a role that the database does not work with the disk directly, but through the OS cache.

@anayrat, hi 👋 Can you help me with this topic (negative `pg_wal_lsn_diff(flush_lsn, replay_lsn)`)?

@artemsafiyulin I think the changes you propose make total sense. Just want to double-check this negative flush/replay _issue_.

ilyam8:
@artemsafiyulin see [this comment](https://github.com/postgres/postgres/blob/240e0dbacd390a8465552e27c5af11f67d747adb/src/include/replication/walsender_private.h#L48-L55)

```c
	/*
	 * The xlog locations that have been written, flushed, and applied by
	 * standby-side. These may be invalid if the standby-side has not offered
	 * values yet.
	 */
	XLogRecPtr	write;
	XLogRecPtr	flush;
	XLogRecPtr	apply;
```

> These may be invalid if the standby-side has not offered values yet.

That explains why flush/apply diff can be negative.

anayrat:
>     * We really need to calculate the difference between each step in order to understand at which step the delay occurred.

Indeed

> *negative: As can be seen from the graphs, we can get negative values for reply_lag. Because at that moment we wrote the data to the OS cache, after which postgres applied the data faster than the OS flushed them from the cache to disk. It plays a role that the database does not work with the disk directly, but through the OS cache.

I have the same interpretation, but I am not sure.

> Or do we just need the diff between pg_last_wal_receive_lsn() and replay_location and call it replication lag size?

It is also my feeling.  The purpose of Netdata is just to give an overview of database health. Not digging deeply to understand performance issue? Maybe one metric for replication lag is enough ?

artemsafiyulin:
> It is also my feeling. The purpose of Netdata is just to give an overview of database health. Not digging deeply to understand performance issue? Maybe one metric for replication lag is enough ?

I didn't add any deeper functionality in this MR. Netdata already counts the same metrics, only it counts not the diff between steps, but the diff between the current wal location and each step.
This MR - is just small fix.

ilyam8:
ok, I agree with the changes. Thanks, @artemsafiyulin.

> Because at that moment we wrote the data to the OS cache, after which postgres applied the data faster than the OS flushed them from the cache to disk. It plays a role that the database does not work with the disk directly, but through the OS cache.

This is probably correct, but since no source (supporting this claim) has been provided, it will take me some time to verify it.

ilyam8:
Not a `c` developer and checking the source code doesn't make shed any light on the issue. It works as it is described in the[ walreceiver.c comment](https://github.com/postgres/postgres/blob/5de94a041ed7a51b571db2030ba87600c7fc6262/src/backend/replication/walreceiver.c#L13-L16)

> As XLOG records are received and flushed to disk, it updates the WalRcv->flushedUpto variable in shared memory, to inform the startup process of how far it can proceed with XLOG replay.

I don't see how any caching is involved.

- in [XLogWalRcvSendReply](https://github.com/postgres/postgres/blob/5de94a041ed7a51b571db2030ba87600c7fc6262/src/backend/replication/walreceiver.c#L1101) we can see what fields are used to create a message that is sent to the primary instance (indicating the slave node current WAL locations).
- flushedUpto (used for replay) is changed to LogstreamResult.Flush in [XLogWalRcvFlush](https://github.com/postgres/postgres/blob/5de94a041ed7a51b571db2030ba87600c7fc6262/src/backend/replication/walreceiver.c#L994) after flushing (`fdatasync()`).

So it is `write()`, `flush()` => wake-up recovery thread that replays data up to `flushedUpto`.

artemsafiyulin:
> Not a `c` developer and checking the source code doesn't make shed any light on the issue. It works as it is described in the[ walreceiver.c comment](https://github.com/postgres/postgres/blob/5de94a041ed7a51b571db2030ba87600c7fc6262/src/backend/replication/walreceiver.c#L13-L16)
> 
> > As XLOG records are received and flushed to disk, it updates the WalRcv->flushedUpto variable in shared memory, to inform the startup process of how far it can proceed with XLOG replay.
> 
> I don't see how any caching is involved.
> 
> * in [XLogWalRcvSendReply](https://github.com/postgres/postgres/blob/5de94a041ed7a51b571db2030ba87600c7fc6262/src/backend/replication/walreceiver.c#L1101) we can see what fields are used to create a message that is sent to the primary instance (indicating the slave node current WAL locations).
> * flushedUpto (used for replay) is changed to LogstreamResult.Flush in [XLogWalRcvFlush](https://github.com/postgres/postgres/blob/5de94a041ed7a51b571db2030ba87600c7fc6262/src/backend/replication/walreceiver.c#L994) after flushing (`fdatasync()`).
> 
> So it is `write()`, `flush()` => wake-up recovery thread that replays data up to `flushedUpto`.

@ilyam8 Thank you, I agree that the comments you specified in the code tell us that a negative lag is impossible. 
But I did another test, where I also generated a replication lag, and also got a negative lag at the "replay" stage (image below)
replay_negative_lag_grafana

I also looked at what netdata shows us at this moment. The first [metric](https://github.com/netdata/go.d.plugin/blob/master/modules/postgres/queries.go#L321-L331) is Replication WAL lag size (which this MR modifies). It currently counts:
- for flush_lag - diff between "pg_current_wal_lsn" and "flush_lsn"
- for replay_lag - diff between "pg_current_wal_lsn" and "flush_lsn". 
The resulting graph is shown in the image below.
replay_negative_lag_netdata_1

The second [metric](https://github.com/netdata/go.d.plugin/blob/master/modules/postgres/queries.go#L341-L342) is Replication WAL lag time. It currently outputs the "flush_lag" and "replay_lag" fields from the pg_stat_replication view. According to the [documentation](https://www.postgresql.org/docs/13/monitoring-stats.html#MONITORING-PG-STAT-REPLICATION-VIEW), the lag in these fields is also calculated between the moment of reset on the master and the moment of flash and replay, respectively, on the replica. The resulting graph is shown in the image below.
replay_negative_lag_netdata_2

As can be seen from both netdata graphs, "replay_lag" is smaller than "flush_lag" in both cases. This is similary to a negative lag. That is, we replay more records than flushed.
This test confirms the correctness of the calculation of the replication lag at each step that I propose in this MR. When viewing this graph, we will not be interested in a small deviation in the negative direction, we will be interested in spikes at certain stages to understand where the replication lag occurs.
And the search for the reason why all three considered methods of fixing the replication lag can show a negative lag, I suggest putting it into a separate discussion and sending this problem to PostgreSQL developers.

ilyam8:
@artemsafiyulin my point is:
 - negative values are not expected (probably wrong, but this is the result of a (brief) check of the source code).
 - we can't explain negative values properly like 

> **positive**: the data is stored on a disk (slave) that is not yet replayed and is not accessible to end users.
> **negative**: how do I interpret it?

- Sooner (likely) or later somebody will open a bug report "Replay delta is negative".

All these considered, I suggest to 0 *delta if the result is < 0 (and add a TODO to debug the issue) - I will add it to the follow-up PR.

ilyam8 (approved):

PR #1075, opened 2023-02-09
Rename module (WMI -> Windows)

thiagoftsm:
Fixes https://github.com/netdata/netdata/issues/14439

This PR is modifying `wmi` collector to `windows monitoring`.

Tests were already done in one of our VMs where charts appear with new 
[contexts](https://github.com/netdata/go.d.plugin/files/10691812/charts.txt):

Screenshot_20230209_013810


modules/windows/collect.go
```diff
@@ -35,7 +35,7 @@ const (
 	collectorNetFrameworkCLRSecurity        = "netframework_clrsecurity"
 )
 
-func (w *WMI) collect() (map[string]int64, error) {
+func (w *WINDOWS) collect() (map[string]int64, error) {
```
ilyam8, line 38: Should be just `Windows`

ilyam8 (changes requested):

modules/windows/collect.go
thiagoftsm: I updated documentation removing `monitoring`.

thiagoftsm:
I convert to draft while we discuss compatibility break.

modules/windows/collect.go
ilyam8: I didn't mean documentation: `func (w *WINDOWS)` => `func (w *Windows)`

thiagoftsm: I pushed another commit renaming it. I was using the same pattern we have in `netdata/netdata`.

PR #1076, opened 2023-02-09
prefer user wmi.conf over windows.conf

ilyam8:
https://github.com/netdata/go.d.plugin/pull/1075 follow-up PR.

We have no logic that handles module renaming. It is the first time we do it, so not sure we need it all. This is just a quick and temporary fix. We need to announce this change in the next Netdata release notes and then remove it.

### DIFF
--- a/agent/setup.go
+++ b/agent/setup.go
@@ -119,6 +119,22 @@ func (a *Agent) buildDiscoveryConf(enabled module.Registry) discovery.Config {
 	}
 
 	for name := range enabled {
+		// TODO: properly handle module renaming
+		// This is just a quick fix for wmi=>windows. We need to prefer user wmi.conf over windows.conf
+		// 2nd part of this fix is in /agent/job/discovery/file/parse.go parseStaticFormat()
+		if name == "windows" {
+			cfgName := "wmi.conf"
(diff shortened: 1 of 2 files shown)


agent/setup.go
```diff
@@ -119,6 +119,23 @@ func (a *Agent) buildDiscoveryConf(enabled module.Registry) discovery.Config {
 	}
 
 	for name := range enabled {
+		// TODO: properly handle module renaming
+		// We need to announce this change in Netdata v1.39.0 release notes and then remove this workaround.
+		// This is just a quick fix for wmi=>windows. We need to prefer user wmi.conf over windows.conf
+		// 2nd part of this fix is in /agent/job/discovery/file/parse.go parseStaticFormat()
+		if name == "windows" {
+			cfgName := "wmi.conf"
```
thiagoftsm, line 127: Why cannot we check both files as temporary?

ilyam8: What do you mean?

- we look for user wmi.conf
- if it doesn't exist, we look for windows.conf (both user and stock)

thiagoftsm: My bad, I make a confusion about `nil` meaning.

thiagoftsm (approved):

PR #1078, opened 2023-02-10
Exchange metrics

thiagoftsm:
This PR is bringing Exchange metrics for Windows module.

Screenshot_20230213_170803

All charts are available as you can see in this [file](https://github.com/netdata/go.d.plugin/files/10724564/charts.txt)


shyamvalsan:
Looks good @thiagoftsm all the metrics seem to be there, can we re-name and re-order as below (and is there a rule to not use upper case? @ilyam8 its a bit weird when acronyms like OWA, LDAP are lower case)

queue -> mail queue
ldap -> LDAP
owa -> OWA
sync -> ActiveSync
rpc -> RPC client access
workload -> workload management
proxy -> HTTP proxy

thiagoftsm:
> Looks good @thiagoftsm all the metrics seem to be there, can we re-name and re-order as below (and is there a rule to not use upper case? @ilyam8 its a bit weird when acronyms like OWA, LDAP are lower case)
> 
> queue -> mail queue ldap -> LDAP owa -> OWA sync -> ActiveSync rpc -> RPC client access workload -> workload management proxy -> HTTP proxy

Yes, we use lowercase internally, and we use dashboard info to convert to upper case.

shyamvalsan (approved):

ilyam8 (approved):

PR #1083, opened 2023-02-13
Fix broken links in .md files

Ancairon:
Related to netdata/learn#1349 
This is a WIP PR intended to fix broken links in markdown files, if there are any issues, where the missing link's info didn't get moved, but deleted, I will add checkbox items to decide what we should do about it.

- The [Contributing](https://github.com/netdata/go.d.plugin/blob/master/README.md?plain=1#:~:text=%23%23%20Contributing,done%20with%20testing.) section in the `README.md` has some old Learn links, and I removed a link that referred to a missing document.
- [The pulsar README](https://github.com/netdata/go.d.plugin/blob/master/modules/pulsar/README.md?plain=1#:~:text=To%20check%20matcher,matcher/README.md).) has a link to [`/pkg/matcher/README.md`](https://github.com/netdata/go.d.plugin/blob/master/pkg/matcher/README.md?plain=1) which we don't include in Learn. I will add it in a new section, to not confuse the current ordering effort, and we can decide then where it should go. (EDIT, this is the case for all the READMEs under `/pkg`)

### DIFF
--- a/pkg/README.md
+++ b/pkg/README.md
@@ -1,15 +1,22 @@
+<!--
+title: "Helper Packages"
+custom_edit_url: "https://github.com/netdata/go.d.plugin/edit/master/pkg/README.md"
+sidebar_label: "Helper Packages"
+learn_status: "Published"
+learn_rel_path: "Developers/External plugins/go.d.plugin"
+-->
+
 # Helper Packages
 
 - if you need IP ranges consider to
-  use [`iprange`](https://github.com/netdata/go.d.plugin/tree/master/pkg/iprange#iprange).
+  use [`iprange`](https://github.com/netdata/go.d.plugin/blob/master/modules/pulsar/README.md).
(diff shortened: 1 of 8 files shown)


pkg/README.md
```diff
@@ -1,15 +1,22 @@
...
+sidebar_label: "Helper Packages"
+learn_status: "Published"
+learn_rel_path: "Developers/External plugins/go.d.plugin/Helper Packages"
+-->
+
 # Helper Packages
 
 - if you need IP ranges consider to
-  use [`iprange`](https://github.com/netdata/go.d.plugin/tree/master/pkg/iprange#iprange).
+  use [`iprange`](https://github.com/netdata/go.d.plugin/blob/master/modules/pulsar/README.md).
```
ilyam8, line 12:
```suggestion
  use [`iprange`](https://github.com/netdata/go.d.plugin/blob/master/pkg/iprange/README.md).
```

PR #1127, opened 2023-03-17
fix unknown HOST on cleanup

ilyam8:


Ferroin (approved):
Tested locally, works as advertised.

PR #1164, opened 2023-04-25
Windows hyperv

thiagoftsm:
This PR is bringing hyperv to `go.d.plugin`. 

Screenshot_20230425_020745

This PR covers the following metrics:
- [x] Global metrics
- [x] Devices
- [x] Interfaces
- [x] core
- [x] vswitch
- [x] VMs


ilyam8:
I had to change quite a lot:
 - all in/out type metrics (received/sent, incoming/outgoing, reads/writes, etc.) grouped into one chart.
 - moved HyperV metrics into their own section (they were under "Windows") + better grouping (families).

<details>
<summary>screenshot</summary>

Screenshot 2023-05-05 at 01 51 09

</details>


 - removed some metrics because I couldn't understand them.
 - removed CPU metrics for now because it was a no-go as they were (separate chart per CPU state, not measured in percentage). We need to make them in percentage and states should be dimensions.

Also, tried to make a section per VM/vswitch (the same way as cgroups) but decided not to do because vm storage devices and network interfaces have no `vm` label so it is not possible to associate them with vms.

ilyam8 (approved):

thiagoftsm:
Among the modules we worked, this was one of the most undocumented.

PR #1188, opened 2023-05-23
Metrics fixes

ralphm:
This fixes some issues with units, particularly in hyperv metrics that had columns swapped.


Ancairon:
@ralphm thanks for catching this, as Ilya said, metrics.csv is current source of truth, so this is good to go, we can regenerate the yaml files afterwards.

ilyam8 (approved):

PR #1194, opened 2023-06-02
Update collector yamls

Ancairon:


modules/dnsmasq/metadata.yaml
```diff
@@ -1,3 +1,95 @@
...
+          description: An example configuration.
+          folding:
+            title: Example
+            enabled: true
+          data: |
+            jobs:
+              - name: local
+                address: '127.0.0.1:53'
+        - name: Basic example with `protocol` option
+          description: Local server with defined DNS query transport protocol.
```
ilyam8, line 69: I believe `udp` is the default, an example could use non-default.

modules/dnsmasq/metadata.yaml
```diff
@@ -1,3 +1,95 @@
...
+        - name: Basic
+          description: An example configuration.
+          folding:
+            title: Example
+            enabled: true
+          data: |
+            jobs:
+              - name: local
+                address: '127.0.0.1:53'
+        - name: Basic example with `protocol` option
```
ilyam8, line 68:
```suggestion
        # TODO: should use non-default protocol (not `udp`)
        - name: Basic example with `protocol` option
```

ilyam8 (approved):

PR #1197, opened 2023-06-07
Finalize ActiveMQ yaml

Ancairon:
Address TBDs and change some variables to the discussed ones.

### DIFF
--- a/modules/activemq/metadata.yaml
+++ b/modules/activemq/metadata.yaml
@@ -9,11 +9,7 @@ overview:
       This plugin collects queues and topics metrics using ActiveMQ Console API.
 setup:
   prerequisites:
-    list:
-      - title: TBD
-        text: |
-          TBD
-      
+    list: []
   configuration:
     options:
       description: |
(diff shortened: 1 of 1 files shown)


modules/activemq/metadata.yaml
```diff
@@ -157,10 +153,10 @@ metrics:
   folding:
     title: Metrics
     enabled: false
-  description: TBD
+  description: The following sections are presented *by scope*, each scope indicates that there will be a metric for every scope (e.g database, interface etc). Global scope means that the metrics are collector-wide.
```
ilyam8, line 156:
```suggestion
  description: |
    Metrics grouped by *scope*.
    
    The scope defines the instance that the metric belongs to. An instance is uniquely identified by a set of labels.
```

modules/activemq/metadata.yaml
```diff
@@ -157,10 +153,10 @@ metrics:
...
   scope:
     - name: global
-      description: TBD
+      description: ""
```
ilyam8, line 159:
```suggestion
      description: |
        Global scope means that the metrics apply to the entire monitored application.
```

ilyam8 (approved):

PR #1207, opened 2023-06-15
Update readmes and fix some yaml inconsistencies

Ancairon:
@ilyam8 please check dnsmasq_dhcp and if you agree with the scope description let me know to make the readme too.

### DIFF
--- a/modules/dnsdist/README.md
+++ b/modules/dnsdist/README.md
@@ -1,61 +1,126 @@
-<!--
-title: "DNSdist monitoring with Netdata"
-description: "Monitor the health and performance of DNSdist load balancers with zero configuration, per-second metric granularity, and interactive visualizations."
-custom_edit_url: "https://github.com/netdata/go.d.plugin/edit/master/modules/dnsdist/README.md"
-sidebar_label: "DNSdist"
-learn_status: "Published"
-learn_topic_type: "References"
-learn_rel_path: "Integrations/Monitor/Networking"
--->
-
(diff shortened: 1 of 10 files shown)


modules/dnsdist/README.md
```diff
@@ -118,23 +118,32 @@ jobs:
       X-API-Key: your-api-key # static pre-shared authentication key for access to the REST API (api-key).
 ```
 
+</details>
+
 ##### Multi-instance
 
-When you are defining more than one jobs, you must be careful to use different job names, to not override each other.
+> **Note**: When you define multiple jobs, their names must be unique.
```
ilyam8, line 125: This note should go to Configuration->file (we can hardcode it in the template, no need to have it in meta)

ilyam8 (approved):

modules/dnsmasq_dhcp/metadata.yaml
```diff
@@ -21,55 +23,47 @@ setup:
...
-          folding:
-            title: Example
-            enabled: true
           data: |
             jobs:
               - name: dnsmasq_dhcp
                 leases_path: /var/lib/misc/dnsmasq.leases
                 conf_path: /etc/dnsmasq.conf
                 conf_dir: /etc/dnsmasq.d
-        - name: Multi-instance
```
ilyam8, line 60: Mutli-instance for this collector has no sense, it can't be several instances of Dnsmasq on the same host

PR #1232, opened 2023-07-07
Update cassandra yaml

Ancairon:
Make cassandra yaml into new structure, mainly to have an example, and to get a feeling for how it is filling it in


modules/cassandra/metadata.yaml
```diff
@@ -168,142 +212,192 @@ metrics:
...
         - name: cassandra.storage_exceptions_rate
+          availability: []
           description: Storage exceptions rate
           unit: exceptions/s
+          chart_type: line
           dimensions:
             - name: storage
     - name: thread pool
-      description: Metrics related to Cassandra's thread pools. Each thread pool provides its own set of the following metrics.
+      description: ''
```
ilyam8, line 371: ?

Ancairon: well, just forgot that, will add it back in

modules/cassandra/metadata.yaml
```diff
@@ -168,142 +212,192 @@ metrics:
...
             - name: pending
         - name: cassandra.thread_pool_blocked_tasks_count
+          availability: []
           description: Blocked tasks
           unit: tasks
+          chart_type: line
           dimensions:
             - name: blocked
         - name: cassandra.thread_pool_blocked_tasks_rate
+          availability: []
```
ilyam8, line 398: I think `availability` can be optional at this level.

PR #1241, opened 2023-07-17
Yaml updates

Ancairon:
will mark ready when bugfix is ran for the other ready yamls.

### DIFF
--- a/modules/consul/metadata.yaml
+++ b/modules/consul/metadata.yaml
@@ -244,7 +244,7 @@ metrics:
   folding:
     title: Metrics
     enabled: false
-  description: ""
+  description: Depending on the [mode](https://developer.hashicorp.com/consul/docs/install/glossary#agent), the collector collects a different number of metrics.
   availability: []
   scopes:
     - name: global
--- a/modules/consul/metadata.yaml
+++ b/modules/consul/metadata.yaml
@@ -244,7 +244,8 @@ metrics:
   folding:
     title: Metrics
     enabled: false
-  description: Depending on the [mode](https://developer.hashicorp.com/consul/docs/install/glossary#agent), the collector collects a different number of metrics.
+  description: |
+    Depending on the [mode](https://developer.hashicorp.com/consul/docs/install/glossary#agent), the collector collects a different number of metrics.
   availability: []
   scopes:
     - name: global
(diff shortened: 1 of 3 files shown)


modules/elasticsearch/metadata.yaml
```diff
@@ -213,7 +213,32 @@ setup:
 troubleshooting:
   problems:
     list: []
-alerts: []
+alerts:
+  - name: elasticsearch_node_indices_search_time_query
+    link: 
+      https://github.com/netdata/netdata/blob/master/health/health.d/elasticsearch.conf
```
ilyam8, line 219: I think it is better to have them on the same line (link: _link_)


What link do we need here? To GH HTML page or [raw content](https://raw.githubusercontent.com/netdata/netdata/master/health/health.d/elasticsearch.conf)?

modules/elasticsearch/metadata.yaml
```diff
@@ -598,4 +623,4 @@ metrics:
           unit: bytes
           chart_type: line
           dimensions:
-            - name: store_size
+            - name: store_size
```
ilyam8, line 626:
```suggestion
            - name: store_size

```

modules/elasticsearch/metadata.yaml
Ancairon: agree, let me format the files?

Ancairon: Yes, let me format the files. VScode does this sometimes, I see it is doing it for labels further down below. Will add `|` manually there if we don't like how this looks.

I think HTML page, not raw

modules/consul/metadata.yaml
```diff
@@ -1,120 +1,161 @@
...
+      - data-collection.service-discovery-registry
+    icon_filename: "consul.png"
+  related_resources:
+    integrations:
+      list: []
+  info_provided_to_referring_integrations:
+    description: ""
+  keywords:
+    - service networking platform
+    - hashicorp
```
ilyam8, line 18: Do we need to add the app name as the keyword (add `consul` here)?

ilyam8 (approved):

modules/consul/metadata.yaml
Ancairon: I'd say no, probably it will be searchable by its name, maybe @kapantzak can confirm that we don't need consul as keyword because that is the name

PR #1248, opened 2023-07-18
Attempt to add chart_type, and make most yamls pass schema

Ancairon:
chart type waas added, multiline strings have `\n`s now, but should work.

Some keys were added, but I can't control ordering in yaml, so they are not sorted. 

In yamls having only metrics I added the empty template to make them pass the schema.


ilyam8 (changes requested):
too many `\n`, sorry, wontfix

PR #1289, opened 2023-08-01
Integration categorizations

Ancairon:


modules/prometheus/metadata.yaml
```diff
@@ -4791,7 +4791,7 @@ modules:
         link: https://github.com/wikimedia/operations-software-hhvm_exporter
         icon_filename: hhvm.svg
         categories:
-          - data-collection.web-servers-and-web-proxies
+          - data-collection.containers-and-vms
```
ilyam8, line 4794: Doesn't looks like containers-and-vms to me. Why?

Ancairon: was instructed to do so, and I made this PR so everyone could add here, which wasn't the case after all. Isn't this https://hhvm.com/ ?

ilyam8: Yes, it is.

> HHVM is an [open-source](http://github.com/facebook/hhvm) virtual machine designed for executing programs written in [Hack](http://hacklang.org/)

Doesn't look like it is our containers-and-vms category.

PR #1292, opened 2023-08-01
Update prometheus metadata.yaml

sashwathn:
Updated icon for DNSBL

PR #1296, opened 2023-08-01
Update metadata.yaml

sashwathn:
Moved Crowdsec, OpenVAS and Clam related integrations to data-collection.security-systems category


ilyam8:
Fixed in #1292

PR #1298, opened 2023-08-01
Update metadata.yaml

sashwathn:
Updated categories for Mesos and Slurm to data-collection.ci-cd-systems


ilyam8:
Fixed in #1292

PR #1253, opened 2023-07-20
add prometheus integrations to multi_metadata.yaml

shyamvalsan:
Keywords still to be updated.

PR #1303, opened 2023-08-01
Change Prometheus Exporter Categorizations

ralphm:


ilyam8 (approved):

ilyam8:
cc @shyamvalsan (you added initial categories in #1253)

shyamvalsan:
@ralphm @ilyam8 The base categories were just using what existed already on the website. And we need to fix stuff for sure.

- Generic data collection is intended to cover prometheus, statsd, snmp scenarios, where it can be used to collect data from a variety of sources.
- GitHub and GitHub API should both be somewhere else
- If we need a catch all, it should be "Other"
- I think @sashwathn already moved MQTT blackbox to message-brokers, PR pending
- Obervability is just about monitoring observability services
- Kubernetes vs Containers & VMs (I think the ask for a separate category for Kubernetes is from @ktsaou, we're just following the precedence from the website here)
- The notifications category has been renamed by @sashwathn to "Notifications and Incident Management" - not sure Discourse fits in either way

PR #1305, opened 2023-08-01
Update metadata.yaml

sashwathn:
Updated the categories for various integrations.

### DIFF
--- a/modules/snmp/metadata.yaml
+++ b/modules/snmp/metadata.yaml
@@ -8,7 +8,7 @@ modules:
         link: ""
         icon_filename: snmp.png
         categories:
-          - data-collection.snmp-and-networked-devices
+          - data-collection.generic-data-collection
       keywords:
         - snmp
       related_resources:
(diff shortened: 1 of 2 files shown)


shyamvalsan (approved):

modules/snmp/metadata.yaml
```diff
@@ -8,7 +8,7 @@ modules:
         link: ""
         icon_filename: snmp.png
         categories:
-          - data-collection.snmp-and-networked-devices
```
ilyam8, line 11: I think having SNMP in a separate category (snmp/network devices) made more sense but ok

sashwathn: Having just a single integration in the category did not make sense when it can anyways be searched.. 
We may reintroduce this category based on user feedback.

cc: @ralphm @Ferroin

ilyam8: i wish I was a user and can give feedback 😢

sashwathn: I am creating a follow up issue for improving integrations with all of our feedback - as users of Netdata.. 👍

PR #1312, opened 2023-08-19
added CrowdSec Prometheus metrics URL

luisj1983:
Added CrowdSec metrics URL to prometheus.conf.

### DIFF
--- a/config/go.d/prometheus.conf
+++ b/config/go.d/prometheus.conf
@@ -1546,3 +1546,5 @@ jobs:
   - name: besu_local
     url: '127.0.0.1:9545'
     expected_prefix: 'besu_'
+  - name: crowdsec_local
+    url: http://127.0.0.1:6060/metrics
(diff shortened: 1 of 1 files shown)


config/go.d/prometheus.conf
```diff
@@ -1546,3 +1546,5 @@ jobs:
   - name: besu_local
     url: '127.0.0.1:9545'
     expected_prefix: 'besu_'
+  - name: crowdsec_local
+    url: http://127.0.0.1:6060/metrics
```
ilyam8, line 1550: Can you please add `expected_prefix`?

luisj1983: Is there a syntax for multiple prefixes?

ilyam8: @luisj1983 we need just one that would be unique for crowdsec. It is ok if this exporter has metrics with other prefixes.

luisj1983: @ilyam8 OK, done. Anything else I need to do prior to a merge?

ilyam8 (approved):

config/go.d/prometheus.conf
ilyam8: No, thanks

ilyam8:
Thanks, @luisj1983.

PR #1330, opened 2023-09-08
Create generate-integrations-docs.yml

Ancairon:
This should be merged after https://github.com/netdata/netdata/pull/15893 is merged and confirmed that it runs as expected. Until that is merged, I label this as WIP so it doesn't create confusion.

The action fetches `integrations.js` and `gen_docs_integrations.py` and creates integration docs to then use on Learn.


.github/workflows/generate-integrations-docs.yml
```diff
@@ -0,0 +1,49 @@
+---
+  # CI workflow used to generate documentation from integrations/integrations.js.
+
+name: Generate Go Integrations Documentation
+on:
+  schedule:
+    - cron: '0 10 * * *'
+  workflow_dispatch: null
```
ilyam8, line 8: Why cron? Should it be 
- pushes to master + 
- some files filter
 
?

Ancairon: the ideal would be to track changes on the integrations.js, we don't want to change readmes when yaml files change, as the integrations.js will be the same :thinking: 

Can we track file changes from other repos?

ilyam8: Ah, I see. As it is now this workflow should be triggered by the one in netdata/netdata (which in 99% cases will do nothing) or cron.

ilyam8: Probably it is better to not add this workflow in netdata/go.d.plugin but have it in netdata/netdata and let it make a PR in netdata/go.d.plugin if needed. What do you think?

Ancairon: The script atm is designed to run in both repos, I haven't tested it to run with go checked out in the netdata repo :thinking: 

How would we then PR the files over to here?

I thought the simplest would be to run it with cron, and then if we merge the integrations PR in the other repo, manually run it.

@Ferroin there is no trigger to refer to a merged PR on netdata/netdata eh?
If there was, we would use that

Ferroin: For updating things in the agent repo, we should absolutely just run it in the same workflow that is generating the integrations.js file.

For updating things here, add a `workflow_dilspatch` trigger on this workflow, and I’ll add a workflow in the agent repo to trigger it whenever the `integrations.js` file changes on the master branch.

.github/workflows/generate-integrations-docs.yml
```diff
@@ -0,0 +1,47 @@
+---
+  # CI workflow used to generate documentation from integrations/integrations.js.
+
+name: Generate Go Integrations Documentation
+on: 
+  workflow_dispatch:
```
Ferroin, line 6:
```suggestion
  workflow_dispatch: null
```

This is not strictly required by GHA, but it makes a number of YAML linting tools happier, and is consistent with what we do in workflows in the agent repo when a trigger needs no other configuration options.

Ferroin (approved):

PR #1334, opened 2023-09-15
openvpn_status_log: Do not count "UNDEF" users

vpnable:
`UNDEF` clients should not be counted, [as it means](https://forums.openvpn.net/viewtopic.php?p=89781#p89781) the connection is not established yet. If you don't use `tls-auth` in the openvpn server config, [random UDP packets or DoS](https://sourceforge.net/p/openvpn/mailman/openvpn-users/thread/alpine.LRH.1.10.0810221455570.5642%40rhlx09.hs-esslingen.de/#msg20611785) could also create a lot of `UNDEF` users in the status files.

The old [python-based openvpn plugin handled this](https://github.com/netdata/netdata/pull/3734) as well and the current `openvpn` module seems to have [special handling](https://github.com/netdata/go.d.plugin/blob/5fa3ff0de80f2a9055c1d1983e0cf1aada48dfbf/modules/openvpn/collect.go#L55) of `UNDEF` too.

### DIFF
--- a/modules/openvpn_status_log/collect.go
+++ b/modules/openvpn_status_log/collect.go
@@ -57,7 +57,7 @@ func (o *OpenVPNStatusLog) collectUsers(mx map[string]int64, clients []clientInf
 func numOfClients(clients []clientInfo) int64 {
 	var num int64
 	for _, v := range clients {
-		if v.commonName != "" {
+		if v.commonName != "" && v.commonName != "UNDEF" {
 			num++
 		}
 	}


ilyam8:
Thanks, @vpnable. Regarding the special handling of UNDEF users in `openvpn` - I guess we need to add it to `openvpn_status_log` too.

- `openvpn` collectUsers()

https://github.com/netdata/go.d.plugin/blob/5fa3ff0de80f2a9055c1d1983e0cf1aada48dfbf/modules/openvpn/collect.go#L54-L58

- `openvpn_status_log` collectUsers()

https://github.com/netdata/go.d.plugin/blob/5fa3ff0de80f2a9055c1d1983e0cf1aada48dfbf/modules/openvpn_status_log/collect.go#L40-L41


To be honest I don't remember why I did this special handling. I wish I would add more comments 😞

ilyam8 (approved):

PR #1345, opened 2023-09-26
Adding new lines to fix details tag

hugovalente-pm:
image

### DIFF
--- a/modules/vcsa/metadata.yaml
+++ b/modules/vcsa/metadata.yaml
@@ -253,8 +253,10 @@ modules:
             |   red   | Security patches might be available.                 |
             |  gray   | An error retrieving information on software updates. |
             | unknown | Collector failed to decode status.                   |
+
             
             </details>
+            
           labels: []
           metrics:
             - name: vcsa.system_health_status


ilyam8:
@hugovalente-pm what is the problem? I think the syntax is ok in the metadata. Probably gen script should be adjusted not meta.

hugovalente-pm:
while working on github issues/comments and Learn I had come across places where not having an _adequate_ split between the html element and the next piece of markdown code it was causing for the html attributes to not work properly. I thought the issue was the same here

if you think this is the gen script let's ping @Ferroin to check

ilyam8:
Yeah, I suspect there is a problem in the gen script because there is an empty line between the end of the table and `</details>`

Screenshot 2023-09-26 at 18 07 33

ilyam8:
@Ancairon ping

Ancairon:
This is interesting, I would also add a newline at line 222-223

ilyam8:
What is interesting? 🤔 I don't get why we need to add newlines in that meta file, it looks ok to me.

Ancairon:
there is no need to, I agree, let me check what comes up in the big PR in markdown form and come back here

Ancairon:
see https://github.com/netdata/go.d.plugin/blob/integrations-docs/modules/vcsa/integrations/vcenter_server_appliance.md

that's why I said it is interesting, this is a problem of the app not being able to properly parse markdown, that's my guess

ilyam8:
@Ancairon asking you because we could have this problem in multiple places and we just don't know about that.

Ancairon:
which in this case is a @kapantzak to make sure

ilyam8:
Closing because it is not meta problem

PR #1346, opened 2023-09-26
Do not stop collecting mysql statistics because of timeout

chpfm:
Sometimes under heavy load on a read replica `SHOW ALL SLAVES STATUS` freeze and does not complete in time. This breaks the statistics collection. If you don't mind, it would be great if timeout errors were not considered so critical

### DIFF
--- a/modules/mysql/collect.go
+++ b/modules/mysql/collect.go
@@ -5,6 +5,7 @@ package mysql
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


modules/mysql/collect.go
```diff
@@ -68,17 +68,25 @@ func (m *MySQL) collect() (map[string]int64, error) {
...
 		if err := m.collectSlaveStatus(mx); err != nil {
-			m.Errorf("error on collecting slave status: %v", err)
-			// TODO: shouldn't disable on any error
-			m.doSlaveStatus = false
+			if errors.Is(err, context.DeadlineExceeded) {
+				m.Warning("Deadline exceeded during slave status collection")
+			} else {
+				m.Errorf("error on collecting slave status: %v", err)
+				// TODO: shouldn't disable on any error
+				m.doSlaveStatus = false
```
ilyam8, line 76:
```suggestion
				m.doSlaveStatus = !errors.Is(err, context.DeadlineExceeded)
```

Probably just this + change Errorf to Warningf. What do you think?

ilyam8 (approved):

ilyam8 (approved):

ilyam8:
Thanks, @chpfm!

PR #1363, opened 2023-10-06
Fix newline in table

Ancairon:
If we use `|` inside description fields that go into tables, and then add a newline in them, then the markdown gets messed up.


ilyam8:
Will be fixed in the generation script

PR #1367, opened 2023-10-09
Update modules/prometheus/metadata.yaml to not have conflicting names with native collectors

Ancairon:
related discussion: https://netdata-cloud.slack.com/archives/CLVTYCQBD/p1696254773237659

Essentially we don't want to have two integrations named the same, and going to the same category, as it messes up Learn, and also confuses the user imo.

### DIFF
--- a/modules/prometheus/metadata.yaml
+++ b/modules/prometheus/metadata.yaml
@@ -2323,7 +2323,7 @@ modules:
       most_popular: false
       community: true
       monitored_instance:
-        name: OpenLDAP
+        name: OpenLDAP (community)
         link: https://github.com/tomcz/openldap_exporter
         icon_filename: openldap.svg
         categories:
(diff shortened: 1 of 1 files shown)


ilyam8 (approved):

ilyam8:
Seems a bit lame because it will look like other Prometheus are not community.

Ancairon:
We got the badge on Learn and the banner in the app, do you want to add a `(Prometheus)` suffix?

ilyam8:
I know that we have a badge, just saying that some collectors will have ` (Community)` in their names and some don't.

Ancairon:
but everything would have the suffix, and the badge, it would look ugly, this because the same category has two same-named integrations

ilyam8:
Why all when only community? And you can add less verbose version e.g. `(C)`.

- Memcached
- Memcached (C)

PR #1390, opened 2023-10-26
mysql: Use Performance Schema processlist implementation if available

vobruba-martin:
INFORMATION_SCHEMA.PROCESSLIST is deprecated and subject to removal in a future MySQL release. As such, the implementation of SHOW PROCESSLIST which uses that table is also deprecated.
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-35.html#mysqld-8-0-35-deprecation-removal

The new Performace Schema implementation is available since 8.0.22. https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-22.html#mysqld-8-0-22-performance-schema

### DIFF
--- a/modules/mysql/collect_process_list.go
+++ b/modules/mysql/collect_process_list.go
@@ -2,6 +2,10 @@
 
 package mysql
 
+import (
+	"github.com/blang/semver/v4"
+)
+
 // Table Schema:
 // (MariaDB) https://mariadb.com/kb/en/information-schema-processlist-table/
 // (MySql) https://dev.mysql.com/doc/refman/5.7/en/information-schema-processlist-table.html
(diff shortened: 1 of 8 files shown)


modules/mysql/collect_process_list.go
```diff
@@ -19,8 +23,30 @@ ORDER BY
...
+  AND info NOT LIKE '%PROCESSLIST%' 
+ORDER BY 
+  time;`
+)
+
 func (m *MySQL) collectProcessListStatistics(mx map[string]int64) error {
-	q := queryShowProcessList
+	var q string
+	mysqlMinVer := semver.Version{Major: 8, Minor: 0, Patch: 22}
+	if !m.isMariaDB && m.version.GTE(mysqlMinVer) && m.varPerformanceSchema != "OFF" {
```
ilyam8, line 45:
```suggestion
	if !m.isMariaDB && m.version.GTE(mysqlMinVer) && m.varPerformanceSchema == "ON" {
```

What do you think?

ilyam8 (approved):

ilyam8:
Thanks, @vobruba-martin!

vobruba-martin:
@ilyam8 Thanks for the review!

PR #1394, opened 2023-10-26
Set required golang to 1.21

k0ste:
Fixes netdata/netdata#16278

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/netdata/go.d.plugin
 
-go 1.20
+go 1.21
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.0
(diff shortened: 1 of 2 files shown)


ilyam8:
Thanks, @k0ste!

k0ste:
Added another commit, Run `go get -u=patch ./...` command for major version changes. Seems this also needed for golang

ilyam8:
@k0ste please undo the last commit, it will break the plugin in a lot of places (e.g. won't compile on FreeBSD because wgctrl - cross-compilation no longer works with latest). I have dependabot setup and I update direct dependencies, if there is something is not up to date - it breaks the plugin or requires additional actions (e.g. regenerate generated code after updating radius).

k0ste:
Understood, done. Do I need to do something for fix the CI pipeline issue?

ilyam8:
@k0ste rebase onto the master, mysql testes were broken in master, just pushed a commit to fix them.

k0ste:
Rebased

PR #1434, opened 2023-12-06
Vendor archives

arkamar:
Would it be possible to release vendor archives for new versions? This is how it can be done. I tried to integrate it to a current workflow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ mocks/springboot2/.gradle/
 mocks/tmp/*
 !mocks/tmp/.gitkeep
 .DS_Store
+/vendor
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ build: clean ## Build package
 
 .PHONY: clean
 clean:
-	rm -rf bin
+	rm -rf bin vendor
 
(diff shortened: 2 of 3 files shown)


ilyam8:
Hey, @arkamar. Thanks for the contribution! Can you clarify why this is needed?

arkamar:
I'm working on making an official Gentoo package for the netdata `go.d.plugin`. The patch I've sent helps me as a packager to meet Gentoo's packaging rules, which require separate downloads for dependencies (the package is built in a sandbox without internet connection). Previously, we listed dependencies based on the `go.sum` file, but this method is now deprecated in Gentoo. We use vendor files instead (see [Writing Go Ebuilds - Packaging the dependencies](https://wiki.gentoo.org/wiki/Writing_go_Ebuilds#Packaging_the_dependencies)). Without a vendor archive from upstream, I'd have to manually create one for each release. Having an upstream vendor archive makes maintaining the package easier for me, and could help any other user or distribution that needs pre-vendored dependencies.

ilyam8 (approved):

arkamar:
Thank you for merging it!

PR #1444, opened 2023-12-15
upsd: use "ups.realpower" if available for "ups_load_usage" chart

ilyam8:
Fixes: netdata/netdata#16619


etienn01:
@ilyam8 not sure but I think you may have forgotten to add `writeVar(mx, ups, varUpsRealPower)` in https://github.com/ilyam8/go.d.plugin/blob/master/modules/upsd/collect.go#L87

PR #1472, opened 2024-02-05
fix links that were missed

Ancairon:
sed doesn't seem to have picked these up in my local run, did it with vscode search and replace

we need to do the same in netdata/netdata


tkatsoulas:
did you check for all the moved files until https://github.com/netdata/netdata/commit/0a6b7315a5e0bb0a0053c0e85391a204011fb8f1?

tkatsoulas (approved):

